PR #13, opened 2021-11-12
Add metric forecast example

mkam039:
Adds an example on how to run a forecast on a model and send the result to slack.


docs/metric-forecast.md
```diff
@@ -1 +1,119 @@
...
+See [slack example](slack-example.md), for instructions on how to set up a minimal Slack bot.
+
+This example is built for a model that has two columns: `y` and `ds`, where `y` is a metric measure and `ds` is a timestamp. The metric that we look at is Agent Wait Time in minutes.
+
+## Meta tag
+In a `schema.yml` file, within a target model, a meta tag should be added in order to connect the model to fal:
+```yaml
+    meta:
+      fal:
+        scripts: ["path_to_fal_script.py"]
```
burkaygur, line 13: Should we have this with the `-` notation? Makes it a bit easier to read and type out

PR #34, opened 2021-11-18
Better usage of click for help flag

chamini2:
```
❯ fal --version
fal, version 0.1.0

❯ fal --help
Usage: fal [OPTIONS] COMMAND [ARGS]...

Options:
  --version  Show the version and exit.
  --help     Show this message and exit.

Commands:
  run
  
❯ fal run --help
Usage: fal run [OPTIONS]

Options:
  --project-dir PATH   Directory to look for dbt_project.yml.
  --profiles-dir PATH  Directory to look for profiles.yml.
  --keyword TEXT       Property in meta to look for fal configurations.
  --all                Only run models that ran in the last dbt run.
  --debug              Display debug logging during execution.
  --help               Show this message and exit.
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,4 +35,4 @@ requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.scripts]
-fal = "fal.cli:run"
+fal = "fal.cli:cli"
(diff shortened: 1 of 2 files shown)


mkam039:
Very nice! Should we add a description that will get printed on `--help`

chamini2:
I think as a next step. For now we can merge and then we come up with the correct message to print.

PR #48, opened 2021-11-22
Run tests on PR

turbo1912:


src/fal/dag.py
```diff
@@ -103,12 +103,11 @@ def get_values(self) -> List[FalScript]:
 
 
 class ScriptGraph:
-    graph: List[FalScript]
-    outgoing: Dict[FalScript, List[FalScript]] = {}
-    incoming: Dict[FalScript, List[FalScript]] = {}
-    ordered_list: List[FalScript] = []
-
```
chamini2, line 110: You can keep these here, right?

src/fal/dag.py
```diff
@@ -103,12 +103,11 @@ def get_values(self) -> List[FalScript]:
...
     def __init__(self, models: List[DbtModel], keyword: str, root: str, graph=[]):
```
chamini2, line 106: make `graph=[]` to `_graph`, because it is not supposed to be passed in general, only for testing

chamini2 (approved):

PR #57, opened 2021-11-26
Add FalDbt class for import and reorganize project classes and building

chamini2:


src/faldbt/project.py
```diff
@@ -65,40 +84,234 @@ def __post_init__(self):
...
+
+        lib.write_target(
+            data,
+            self._manifest.nativeManifest,
+            self.project_dir,
+            self.profiles_dir,
+            target_source,
+        )
+
+    def write_to_firestore(self, df: pd.DataFrame, collection: str, key_column: str):
```
mkam039, line 215: We should add a doc string here.

src/faldbt/project.py
```diff
@@ -65,40 +84,234 @@ def __post_init__(self):
...
+            if profile_cred.type != "bigquery":
+                raise FalGeneralException(
+                    "Could not find GCP credentials in profiles.yml"
+                )
+
+            # HACK: using internal method of Bigquery adapter to mock a Firebase credential
+            cred = credentials.ApplicationDefault()
+            cred._g_credential = BigQueryConnectionManager.get_bigquery_credentials(
+                profile_cred
+            )
```
mkam039, line 254: Nice

src/faldbt/project.py
```diff
@@ -65,40 +84,234 @@ def __post_init__(self):
...
+            data = _firestore_dict_to_document(data=item, key_column=key_column)
+            self.firestore_client.collection(collection).document(str(key)).set(data)
+
+    def _setup_firestore(self):
+        app_name = f"fal-{uuid.uuid4()}"
+        profile_cred = self._config.credentials
+        # Setting projectId from the profiles.yml
+        options = {"projectId": profile_cred.database}
+
+        try:
```
mkam039, line 228: Why don't we first try the profiles.yml credentials?

mkam039 (commented):
Looks great! But it's failing integration tests.

src/faldbt/project.py
chamini2: I thought that it would be ideal this way. You prefer profiles first?

chamini2: It's actually debatable what makes sense to o.

mkam039: Sure. Here's an argument:

A user might have several dbt projects, each with dedicated google credentials that they specify in project-specific profiles. On top of that, they might have default credentials setup for personal use. Such user will have to unset their default credentials in order to make fal run properly, which is not a desirable workflow.

chamini2: And this is assuming a user that would have a dbt project to download the data (bigquery), would normally want to upload it to the same google project. I think that is reasonable.

Maybe a problem is how he could download data (bigquery) from one project and upload it to a different one.

This seems to be not easy to fix, but we could tackle it later.

turbo1912 (approved):

PR #60, opened 2021-12-01
Generate aclhemy table from pandas to get better typing and specify dialect based on adapter type

chamini2:


turbo1912 (approved):

src/faldbt/lib.py
```diff
@@ -163,10 +162,23 @@ def write_target(
...
+
+
+def _alchemy_engine(
+    adapter: adapters_factory.Adapter,
+    target: Union[ParsedModelNode, ParsedSourceDefinition],
+):
+    if adapter.type() == "bigquery":
+        return sqlalchemy.create_engine(f"bigquery://{target.database}")
+    if adapter.type() == "postgres":
+        return sqlalchemy.create_engine("postgresql://")
```
mkam039, line 181: We don't need to specify this?

mkam039 (approved):

src/faldbt/lib.py
chamini2: postgres vs postgresql

PR #64, opened 2021-12-03
Add experimental ordering to integration tests

mkam039:


.github/workflows/integration_tests.yml
```diff
@@ -48,3 +48,7 @@ jobs:
           fal run --profiles-dir .
           echo "*** Checking fal output files ***"
           python check_outputs.py
+          echo "*** Starting fal run with experimental ordering ***"
+          fal run --experimental-ordering --profiles-dir .
```
chamini2, line 52: Wouldn't we want to change the code then so that it changes the ordering a bit?

turbo1912: yea we can keep this to test it doenst error. but we can add more integration tests for the ordering

chamini2:
What are we doing with this PR?

mkam039:
> What are we doing with this PR?

I think we can merge as is so we can make sure the experimental ordering isn't completely broken. And then we think of tests that target ordering in particular.

chamini2:
OK, let's do that. I'll let you merge since it's yours

PR #66, opened 2021-12-03
Add target path as part of the context

turbo1912:


chamini2 (approved):
What about offering all the paths, though?

PR #69, opened 2021-12-06
Upgrade to dbt 1.0.0

chamini2:


src/faldbt/project.py
```diff
@@ -117,12 +117,12 @@ def __init__(self, project_dir: str, profiles_dir: str):
             parse.get_dbt_results(self.project_dir, self._config)
         )
 
-        normalized_source_paths = normalize_directories(
-            project_dir, self._config.source_paths
+        normalized_model_paths = normalize_directories(
+            project_dir, self._config.model_paths
```
turbo1912, line 121: is this changing now and source-paths is called model-paths?

chamini2: yes

chamini2:
The failing error is because `dbt run` behaves differently, it seems. I have to investigate more.

If I do:
```shell
pip install dbt==0.21.1
dbt run
pip uninstall dbt dbt-postgres ...
pip install dbt-postgres=1.0.0
fal run
```

It does not fail, so I guess it could be 1 of:

1. the generated tables are different (I checked, they look the same, maybe I'm missing something)
2. a dependency involved in the conversion to pandas stays in the version that works for 0.21 under the hood and generates the difference

- If it's 1, we have nothing to do; 
- if it's 2, we have to find the dependency and make it a specific version to get consistent behaviour (or we can just leave it like this and ignore the difference)

chamini2:
dbt-postgres was the culprit, thanks @burkaygur

src/faldbt/cp/events/types.py
```diff
@@ -0,0 +1,2869 @@
+# NOTE: COPIED FROM https://github.com/dbt-labs/dbt-core/blob/43edc887f97e359b02b6317a9f91898d3d66652b/core/dbt/events/types.py
```
turbo1912, line 1: is this unreleased code?

chamini2: Only released for 1.0.0, so including it for backwards compatibilty

chamini2: I am now checking the dbt version to just avoid running the code that needs this in case we are pre-1.0.0.

This was necessary for 1.0.0 to run, but did not exist for pre-1.0.0 code. So, I included it and ran it every time. Now I only run it when we are using dbt 1.0.0 under the hood.

turbo1912 (approved):

PR #70, opened 2021-12-07
Add Jupyter Notebook example

chamini2:


docs/write_jupyter_notebook.md
```diff
@@ -0,0 +1,35 @@
+# Example 9: Use dbt from a Jupyter Notebook
+
+We also offer fal as an importable package to load in a Python environment to reference and use your dbt tables in a dynamic way.
+
```
turbo1912, line 4: maybe add a sentence here how you might want to import fal to a jupiter notebook to run some ad-hoc analysis

docs/write_jupyter_notebook.md
```diff
@@ -0,0 +1,35 @@
+# Example 9: Use dbt from a Jupyter Notebook
+
+We also offer fal as an importable package to load in a Python environment to reference and use your dbt tables in a dynamic way.
```
turbo1912, line 3: I dont think you need also here, as this might be the first example people might be reading.

maybe something like;

We offer `fal` as an importable package that allows developers to reference and use dbt tables in a dynamic way.

chamini2: but having `also` conveys that is not the main usage and they would find out more about what is the main usage

turbo1912 (approved):

PR #90, opened 2021-12-24
Change dbt 1.0.0 notice to talk about dbt versions support

chamini2:


README.md
```diff
@@ -207,8 +207,15 @@ tickets = faldbt.ref('stg_zendesk_ticket_data')
 # pandas.DataFrame
 ```
 
-## Working with dbt 1.0.0
-Support for dbt 1.0 is work in progress. If you want to try out an alpha version of `fal` with support for dbt 1.0, see [here](https://github.com/fal-ai/fal/releases/tag/v.0.2.0-alpha).
+# Versions of dbt we support
+
+We currently offer _official_ support the following versions, all in the same codebase:
```
turbo1912, line 212: "Versions of dbt we support" -> "Supported dbt versions"

"all in the same codebase" -> Any extra configuration to work with different dbt versions is not needed, latest `fal` version currently supports all versions

PR #122, opened 2022-02-11
Docusaurus Docsite

turbo1912:
currently only has the read.me content from the repo


chamini2 (approved):
Do you want to keep the `docsite/static/img/tutorial` directory?

chamini2:
Can you update https://github.com/fal-ai/fal/blob/main/.github/workflows/integration_tests.yml#L7 and https://github.com/fal-ai/fal/blob/main/.github/workflows/pytest.yml#L7 to ignore these?

PR #125, opened 2022-02-15
Use argparse

mederka:
Replaces click with argparse

### DIFF
--- a/src/fal/cli.py
+++ b/src/fal/cli.py
@@ -12,100 +12,99 @@ from fal.utils import print_run_info
 from faldbt.project import FalDbt, FalGeneralException, FalProject
 
 
-class FalCli(object):
+def run_fal(argv):
 
-    def __init__(self, argv):
-        parser = argparse.ArgumentParser(
-            description="Run Python scripts on dbt models",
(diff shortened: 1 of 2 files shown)


src/fal/cli.py
```diff
@@ -10,93 +12,103 @@
...
+        # Handle version checking
+        version = pkg_resources.get_distribution('fal').version
+        parser.add_argument("-v",
+                            "--version",
+                            action="version",
+                            version=f"fal {version}",
+                            help="show fal version")
+
+        # Handle commands
+        parser.add_argument("command", help="Subcommand to run")
```
chamini2, line 31: Shouldn't this be used instead? https://docs.python.org/3/library/argparse.html#sub-commands

The current solution seems a little _hacky_

mkam039: Yeah, I looked at that, and decided to try this way in order to make it easier to incorporate future commands. 

We can define another method in this class and it will automatically turn into a working command.

chamini2 (approved):

chamini2 (approved):
Looks great!

PR #139, opened 2022-02-17
Add --disable-logging flag

mkam039:


src/fal/cli.py
```diff
@@ -114,6 +117,12 @@ def run_fal(argv):
         action="store_true",
         help="Display debug logging during execution.",
     )
+    run_parser.add_argument(
+        "--disable-logging",
+        action="store_true",
+        help="Disable logging by fal.",
```
chamini2, line 123: I would remove `by fal` because we also don't log dbt-related stuff.

src/fal/cli.py
```diff
@@ -14,6 +14,9 @@
 from fal.utils import print_run_info
 from faldbt.project import FalDbt, FalGeneralException, FalProject
 
+DBT_VCURRENT = dbt.version.get_installed_version()
+DBT_V1 = dbt.semver.VersionSpecifier.from_version_string("1.0.0")
```
chamini2, line 18: The idea is you would use 
```py
from faldbt.lib import DBT_VCURRENT, DBT_V1
```

chamini2 (changes requested):

chamini2 (approved):

PR #146, opened 2022-02-18
Add script_path

omeroguz45:
added scrip_path var with type pathlib.PosixPath that gives the path of the current path script that is being run

### DIFF
--- a/src/fal/run_scripts.py
+++ b/src/fal/run_scripts.py
@@ -39,6 +39,7 @@ class CurrentTest:
 @dataclass
 class ContextConfig:
     target_path: Path
+    script_path: Path
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


chamini2 (approved):
Is this testable? If yes, could we add a test?

src/fal/run_scripts.py
```diff
@@ -39,6 +39,7 @@ class CurrentTest:
 @dataclass
 class ContextConfig:
     target_path: Path
+    script_path: Path
```
chamini2, line 42: Curious why we would want script path and not the directory instead? Or the `project_dir` even?

chamini2 (changes requested):

src/fal/run_scripts.py
omeroguz45: In the data drift example, I wanted to import the anomaly detection functions to use them in the data drift script, and the only way I could do that was to add the dir in which the functions were to the `sys.path`. I brought this up at a standup and we decided that we would add this. [Here's](https://github.com/fal-ai/fal_dbt_examples/blob/283305fac31213ea1cd2dd81b9852a0fd477c9b7/fal_scripts/data_drift.py#L3) where I used it.

chamini2:
Agreed this is no longer necessary since now we offer imports in scripts: https://github.com/fal-ai/fal/pull/190

Issue #121, opened 2022-02-11
write_to_source needs to escape string values


chamini2:
> some rows that contains statements like this one where created_on > '{{date_start}}' will raise and exeception

```
Running query
CREATE TABLE <my schema destination>.<my destination table> (
        id FLOAT(53), 
        sql TEXT, 
        tables TEXT, 
        validation BOOLEAN
)
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/dbt/clients/jinja.py", line 517, in catch_jinja
    yield
  File "/usr/local/lib/python3.9/site-packages/dbt/clients/jinja.py", line 544, in get_template
    return env.from_string(template_source, globals=ctx)
  File "/usr/local/lib/python3.9/site-packages/jinja2/environment.py", line 941, in from_string
    return cls.from_code(self, self.compile(source), globals, None)
  File "/usr/local/lib/python3.9/site-packages/jinja2/environment.py", line 638, in compile
    self.handle_exception(source=source_hint)
  File "/usr/local/lib/python3.9/site-packages/jinja2/environment.py", line 832, in handle_exception
    reraise(*rewrite_traceback_stack(source=source))
  File "/usr/local/lib/python3.9/site-packages/jinja2/_compat.py", line 28, in reraise
    raise value.with_traceback(tb)
  File "<unknown>", line 2046, in template
jinja2.exceptions.TemplateSyntaxError: expected token 'end of print statement', got 'start'
  line 2046
    and rgq.created_at >= ''{{ date start }}''::date
```

In this case, the `sql` column is for SQL statements, which of course has quotes all around.

PR #150, opened 2022-02-22
Escape jinja braces in row string in write_to_source

mkam039:
Addresses #121


mkam039:
> Could we have a test for this? To make sure this does not happen again

I'll do it in a separate PR updating integration tests

PR #162, opened 2022-02-25
Handle flags present in several commands and subcommands

chamini2:
We had a problem when using flags like `fal --profiles-dir . run` vs `fal run --profiles-dir .`, which should behave the same.

This has a "hack" where we create several `dest` for each of these flag positions, but it is all abstracted in the args file.

### DIFF
--- a/src/fal/cli/args.py
+++ b/src/fal/cli/args.py
@@ -1,5 +1,3 @@
-from dataclasses import dataclass
-from email.policy import default
 from uuid import uuid4
 from typing import List
 import os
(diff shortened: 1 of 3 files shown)


src/fal/cli/args.py
```diff
@@ -1,32 +1,65 @@
...
+from typing import List
 import os
 import argparse
 import pkg_resources
 
 
+LEVEL_FLAGS = {}
+
+
+@dataclass(init=False)
```
chamini2, line 13: does this dataclass do anything?

mkam039 (commented):
Looks interesting. Some clarification would be helpful.

turbo1912 (approved):

PR #164, opened 2022-02-26
Get scripts api should return empty if there is not meta in the model

turbo1912:


mkam039 (approved):

src/faldbt/project.py
```diff
@@ -96,21 +96,29 @@ def __post_init__(self):
     def __hash__(self) -> int:
         return self.unique_id.__hash__()
 
-    def get_script_paths(self, keyword, project_dir, before) -> List[Path]:
-        return normalize_directories(project_dir, self.get_scripts(keyword, before))
+    def get_script_paths(
+        self, keyword: str, project_dir: str, before: bool
+    ) -> List[Path]:
+        if self.get_scripts(keyword, before):
```
chamini2, line 102: Since `get_scripts` now handles this case and returns `[]` itself, and since `normalize_directories` does a `map`, which works well with empty lists. Couldn't we leave this as it was? without the `if-else` check?

turbo1912: yea i think we can

src/faldbt/project.py
```diff
@@ -93,21 +96,26 @@ def __post_init__(self):
...
 
-    def get_script_paths(self, keyword, project_dir, before) -> List[Path]:
+    def get_script_paths(
+        self, keyword: str, project_dir: str, before: bool
+    ) -> List[Path]:
         return normalize_directories(project_dir, self.get_scripts(keyword, before))
 
-    def get_scripts(self, keyword, before) -> List[Path]:
+    def get_scripts(self, keyword: str, before: bool) -> List[Path]:
         # sometimes `scripts` can *be* there and still be None
```
chamini2, line 105: this comment should be over (or next to) the `scripts_node = ...` line

src/faldbt/project.py
```diff
@@ -286,15 +297,22 @@ def _find_features(self) -> List[Feature]:
             for column_name in model.columns.keys():
                 if column_name == model.meta[keyword]["feature_store"]["entity_column"]:
                     continue
-                if column_name == model.meta[keyword]["feature_store"]["timestamp_column"]:
+                if (
+                    column_name
+                    == model.meta[keyword]["feature_store"]["timestamp_column"]
+                ):
```
chamini2, line 303: Maybe we can put `model.meta[keyword]["feature_store"]` in a variable to avoid multiline statements like this

alexsandre1985 (commented):
Good

chamini2 (approved):
the comments are not show stoppers, approving

PR #170, opened 2022-03-03
Add behave for integration tests

mkam039:


.github/workflows/integration_tests.yml
```diff
@@ -57,8 +57,10 @@ jobs:
 
       - name: Install test dependencies
         working-directory: integration_tests/mock
-        run: (test -f requirements.txt && pip install -r requirements.txt) || echo "No requirements.txt"
+        run: |
+          pip install behave
```
chamini2, line 61: should this be a dev dependency to be able to run them locally?

integration_tests/features/environment.py
```diff
@@ -0,0 +1,14 @@
+import os
+
+
+def before_all(context):
+    os.system("mkdir mock/temp")
```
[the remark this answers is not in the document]
chamini2: This makes the `after_all` with `rm` not necessary, since it will be different directories.
```
❯ mktemp -d integration_tests/mock/test_XXXXXX
integration_tests/mock/test_BlEPZP
```

integration_tests/features/steps/fal_steps.py
```diff
@@ -0,0 +1,70 @@
+import os
```
chamini2, line 1: this is very cool

.github/workflows/integration_tests.yml
chamini2: I just saw it is a dev dependency. Do we need to pip install it still?

integration_tests/features/environment.py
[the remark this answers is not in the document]
mkam039: We'd still have to create temp dir if we're not using `/tmp`. I think it's OK the way it is right now. Unless you have strong feelings about it. 😉

.github/workflows/integration_tests.yml
mkam039: We don't install dev dependencies in this workflow and I don't think that we have to.

chamini2 (approved):

PR #172, opened 2022-03-04
Airbyte sync

mkam039:
```
from fal.sync.airbyte import AirbyteClient, airbyte_sync

#Starts a sync job and waits for it to finish
job =  airbyte_sync(
  host="test_host",
  connection_id="test_id")

# Api client
airbyte_client = AirbyteClient(host='test_host', max_retries=0)

# has following methods:
api_client.get_connection_data(connection_id="test_id")
api_client.get_job_status(job_id="test_job_id")
api_client.sync(connection_id="test_id")
api_client.sync_and_wait(connection_id="test_id")
```


turbo1912 (approved):

src/fal/el/airbyte.py
```diff
@@ -0,0 +1,129 @@
...
+    FAILED = "failed"
+    ERROR = "error"
+    INCOMPLETE = "incomplete"
+
+
+@dataclass
+class AirbyteClient:
+    """Airbyte REST API connector class."""
+
+    host: str
```
chamini2, line 26: is there a cloud solution from airbyte we could point to by default?

chamini2 (approved):
It's good

src/fal/el/airbyte.py
mkam039: Airbyte cloud doesn't have API access yet.

PR #181, opened 2022-03-09
Flow end to end

turbo1912:


src/fal/cli/__init__.py
```diff
@@ -34,10 +38,18 @@ def cli(argv=sys.argv):
 
         if parsed.command == "flow":
             if parsed.flow_command == "run":
-                parsed_with_before = argparse.Namespace(**vars(parsed), before=True)
-                fal_run(parsed_with_before)
-                dbt_run(parsed)
-                fal_run(parsed)
+                fal_dbt = create_fal_dbt(parsed)
```
chamini2, line 41: Should we move this code to a `fal_flow_run` function?

src/fal/cli/__init__.py
```diff
@@ -46,3 +58,33 @@ def cli(argv=sys.argv):
                 exclude_count=exclude_count,
                 script_count=script_count,
             )
+
+
+def _before_fal_scripts(node_graph: NodeGraph, plan: ExecutionPlan) -> List[FalScript]:
```
chamini2, line 63: Should we change the method to get `List[str]` instead of `ExecutionPlan` and the caller either passes `plan.before_scripts` or `plan.after_scripts`

src/fal/cli/dbt_runner.py
```diff
@@ -59,7 +59,7 @@ def dbt_run(args):
         command_list += ["--profiles-dir", args.profiles_dir]
 
     if args.select:
-        command_list += ["--select"] + args.select
+        command_list += ["--select"] + models_list
     if args.exclude:
         command_list += ["--exclude"] + args.exclude
```
chamini2, line 64: Should we remove `--exclude` and others selectors? I think these are no longer necessary

tests/test_cli.py
```diff
@@ -68,6 +68,32 @@ def test_flow_run_with_project_dir(capfd):
...
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        shutil.copytree(project_dir, tmp_dir, dirs_exist_ok=True)
+        captured = _run_fal(
+            [
+                # fmt: off
+                "flow", "run",
+                "--project-dir", tmp_dir,
+                "--profiles-dir", profiles_dir,
+                "--select",
+                "test.py+"
```
chamini2, line 81: put this in the same line as `--select`, to better transmit it's the flag arg.

chamini2, line 81: dbt [has prefixes](https://docs.getdbt.com/reference/node-selection/syntax#examples) for non-model selections

> The --select flag accepts one or more arguments. Each argument can be one of:
>
> - a package name
> - a model name
> - a fully-qualified path to a directory of models
> - a selection method (path:, tag:, config:, test_type:, test_name:)

What do you think about `--select script:test.py`. This would enable using scripts that do not have `.py` extension.

tests/test_cli.py
```diff
@@ -68,6 +68,32 @@ def test_flow_run_with_project_dir(capfd):
...
+                # fmt: on
+            ],
+            capfd,
+        )
+
+        assert True
```
chamini2, line 87: remove line?

tests/test_cli.py
```diff
@@ -68,6 +68,32 @@ def test_flow_run_with_project_dir(capfd):
...
+        executing_re = re.compile(
+            r"Executing command: dbt --log-format json run --project-dir [\w\/\-\_]+ --profiles-dir [\w\/\-\_]+tests/mock/mockProfile"
```
chamini2, line 89: no `--select XXX` passed?

tests/test_cli.py
chamini2: I would be OK with the `.py` ending, [dbt does it too](https://docs.getdbt.com/docs/building-a-dbt-project/building-models)

> A model is a select statement. Models are defined in .sql files (typically in your models directory):
>
> - Each .sql file contains one model / select statement
> - The name of the file is used as the model name
> - Models can be nested in subdirectories within the models directory

src/fal/cli/__init__.py
turbo1912: yup makes sense

src/fal/cli/dbt_runner.py
turbo1912: exclude will be necessary right, but the logic behind is not implemented yet.. or are you saying we shouldnt support `--exclude` at all?

chamini2: What I'm saying is the select, exclude logic is no longer done in dbt, we do it. So we just pass the `--select` with the correct values. Right?

Let's say:

```
fal flow run --exclude b --select a+
```

With the given nodes:

```mermaid
  graph LR;
      a-->b;
      a-->c;
      b-->d;
      c-->e;
      c-->d;
  style a fill:green
  style b fill:red
  style c fill:green
  style d fill:green
  style e fill:green
```

You figure out that you don't have to do `b` at all in fal, and then you pass to dbt:

```
dbt run --select a c d e
```

which gives a correct run. 

Notice that the `--exclude` is not necessary for the dbt invocation to do the correct thing.

turbo1912: yea I see what you mean, I was thinking we can pass exclude if its just exclude.

 fal flow run --exclude b 
 
 but even then we probably will calculate the dbt models..

src/fal/cli/flow_runner.py
```diff
@@ -0,0 +1,50 @@
...
+from fal.run_scripts import run_scripts
+from fal.cli.dbt_runner import dbt_run
+from fal.cli.fal_runner import create_fal_dbt
+from fal.cli.selectors import ExecutionPlan
+from fal.fal_script import FalScript
+from fal.node_graph import FalFlowNode, NodeGraph, ScriptNode
+from faldbt.project import FalProject
+
+
+def fal_flow_run(parsed):
```
chamini2, line 11: great

chamini2 (approved):
Awesome

PR #186, opened 2022-03-15
Add threads to FalDbt object

chamini2:


src/faldbt/parse.py
```diff
@@ -27,14 +27,15 @@ def get_dbt_user_config(profiles_dir: str) -> UserConfig:
...
     single_threaded: bool
 
 
 def get_dbt_config(
-    project_dir: str, profiles_dir: str, single_threaded=False
+    project_dir: str, profiles_dir: str, threads: Union[int, None] = None
 ) -> RuntimeConfig:
     # Construct a phony config
-    args = RuntimeArgs(project_dir, profiles_dir, single_threaded)
+    args = RuntimeArgs(project_dir, profiles_dir, threads, False)
```
mkam039, line 38: What does `False` stand for? Can we make it a named argument?

mkam039 (approved):
Looks good

src/faldbt/parse.py
chamini2: Sure thing

PR #189, opened 2022-03-15
Pass threads flag to dbt in fal flow run

chamini2:


mkam039 (approved):

chamini2:
I think we should do this, but we could discuss it.

turbo1912 (approved):

PR #204, opened 2022-03-24
Remove path strings from telemetry props

mkam039:


src/fal/telemetry/telemetry.py
```diff
@@ -330,3 +333,24 @@ def wrapper(*args, **kwargs):
...
     return _log_call
+
+
+def _remove_path_from_props(props: Dict) -> Dict:
+    PATH_STR = "[PATH]"
+
+    def _replace_value(val):
+        if isinstance(val, str):
+            if os.path.exists(val):
+                val = PATH_STR
```
chamini2, line 344: What about keeping the basename?
```
>>> x
'/Users/matteo/Projects/fal/fal/integration_tests'
>>> '[PATH]/' + os.path.basename(x)
'[PATH]/integration_tests'
```

src/fal/telemetry/telemetry.py
```diff
@@ -330,3 +333,24 @@ def wrapper(*args, **kwargs):
...
+        elif isinstance(val, list):
+            new_val = []
+            for item in val:
```
chamini2, line 347: Should you just do
```py
for item in val:
  new_val.append(_replace_value(item))
```
?

chamini2: Can even do a map:
```py
val = list(map(_replace_value, val))
```

chamini2 (commented):
Love the `os.path.exists` solution!

src/fal/telemetry/telemetry.py
mkam039: Recursion in a callback, here we go!

src/fal/telemetry/telemetry.py
mkam039: I'd rather not, since we don't know what the basename might contain. An internal dbt project name might be a secret for example.

PR #209, opened 2022-03-28
Stop running when a user scripts fails

chamini2:
TODO:
- [x] integration tests

### DIFF
--- a/src/fal/cli/flow_runner.py
+++ b/src/fal/cli/flow_runner.py
@@ -1,6 +1,6 @@
 from typing import List, cast, Union
 from fal.run_scripts import raise_for_run_results_failures, run_scripts
-from fal.cli.dbt_runner import dbt_run
+from fal.cli.dbt_runner import dbt_run, raise_for_dbt_run_errors
 from fal.cli.fal_runner import create_fal_dbt
 from fal.cli.selectors import ExecutionPlan
 from fal.fal_script import FalScript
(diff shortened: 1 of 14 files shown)


src/fal/run_scripts.py
```diff
@@ -86,6 +86,14 @@ def _prepare_exec_script(script: FalScript, project: FalProject) -> bool:
     return success
 
 
+def handle_run_results(scripts: List[FalScript], results: List[bool]):
```
mkam039, line 89: Why don't we incorporate some of this into `run_scripts`? Maybe we can do mapping/filtering there and just raise an Error in a runner.

mkam039 (commented):
Nice solution. Have a comment about implementation.

src/fal/run_scripts.py
chamini2: Yes, you are right. What if we also raise in the run_scripts for now? Just make it do everything about a script and later we change if we need to. What do you think?

mkam039: Wouldn't this stop the entire fal run? Ideally we want to finish running everything we can and then raise an error.

chamini2: I think we are OK like this for now, and we eventually will move to printing dbt-style logging like

```
Concurrency: 4 threads (target='dev')
1 of 2 START global script scripts/a.py
2 of 2 START model_a script scripts/a.py
2 of 2 OK model_a script scripts/a.py
1 of 2 OK global script scripts/a.py
```

Or something like that

src/fal/cli/flow_runner.py
```diff
@@ -14,12 +14,16 @@ def fal_flow_run(parsed):
     node_graph = NodeGraph.from_fal_dbt(fal_dbt)
     execution_plan = ExecutionPlan.create_plan_from_graph(parsed, node_graph)
 
-    run_scripts(_id_to_fal_scripts(node_graph, execution_plan.before_scripts), project)
+    before_scripts = _id_to_fal_scripts(node_graph, execution_plan.before_scripts)
+    results = run_scripts(before_scripts, project)
+    handle_run_results(before_scripts, results)
```
turbo1912, line 19: let's call this something like `throw_if_failed`

otherwise I am good with this as discussed on Tuesday

turbo1912 (approved):

turbo1912:
I approved but looks like tests are failing

turbo1912 (approved):

PR #213, opened 2022-03-30
Better dbt seeding in tests

turbo1912:


integration_tests/features/environment.py
```diff
@@ -13,4 +14,5 @@ def after_all(context):
 
 
 def after_scenario(context, scenario):
-    os.system("rm -rf mock/temp/*")
+    if not hasattr(context, "temp_dir"):
```
chamini2, line 17: should this be `if hasattr`?

integration_tests/projects/001_flow_run_with_selectors/data/time_series.csv
```diff
@@ -1,4 +1,4 @@
-id,y,ds
```
chamini2, line 1: why?

integration_tests/features/environment.py
```diff
@@ -2,8 +2,9 @@
 
 
 def before_all(context):
-    os.system("mkdir mock/temp")
-    os.system("dbt seed --profiles-dir mock --project-dir mock")
+    if not hasattr(context, "temp_dir"):
+        os.system("mkdir mock/temp")
+        os.system("dbt seed --profiles-dir mock --project-dir mock")
```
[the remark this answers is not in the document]
turbo1912: yea we dont want any of these things in the `environment.py` but we need them for the older tests. .

integration_tests/features/environment.py
turbo1912: so this is an indicator if the tests run with the new style or the old style, we dont want to do this if it ran with the new style..

PR #214, opened 2022-03-30
Add jaffle shop project to integration tests

turbo1912:


integration_tests/features/flow_run_with_jaffle_shop.feature
```diff
@@ -0,0 +1,14 @@
+Feature: `flow run` command
+  Background: Project Setup
+    Given the project 002_jaffle_shop
+    When the data is seeded
```
mkam039, line 4: Should we seed data before individual features or before all features? Right now we're doing both.

turbo1912: do we seed for individual features? we shouldn't be doing that.. we should seed only when we tell the test to seed in this case once for all the tests in the file. 

are you referring to this? 

https://github.com/fal-ai/fal/blob/main/integration_tests/features/environment.py#L8

I added this check here `if not hasattr(context, "temp_dir"):` so that newer style tests dont run this; i thought this worked but maybe doesn't?

mkam039: In my case, it didn't prevent the environment seed from running. Let's let it be for now and write a ticket about adapting old tests to newer style.

mkam039 (approved):

PR #219, opened 2022-04-01
Handle dbt defer flag

mkam039:
Makes a `--defer` work.

```
fal flow run --defer --state my_state_dir
```

### DIFF
--- a/src/fal/cli/args.py
+++ b/src/fal/cli/args.py
@@ -73,6 +73,13 @@ def _build_dbt_common_options(parser: argparse.ArgumentParser):
         dest=_flag_level("profiles_dir"),
     )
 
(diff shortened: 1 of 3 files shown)


chamini2 (approved):

src/fal/cli/args.py
```diff
@@ -73,6 +73,13 @@ def _build_dbt_common_options(parser: argparse.ArgumentParser):
         dest=_flag_level("profiles_dir"),
     )
 
+    parser.add_argument(
```
turbo1912, line 76: why did we add this to the common options but not the state and thread?

mkam039: I guess state could have gone here.

PR #221, opened 2022-04-01
Record run results as fal results after every dbt run

turbo1912:


src/fal/cli/dbt_runner.py
```diff
@@ -66,13 +69,18 @@ def dbt_run(args, models_list: List[str]):
...
-    command_list = list(map(str, command_list)) # make sure all are strings before joining for printing
+    command_list = list(
+        map(str, command_list)
+    )  # make sure all are strings before joining for printing
 
     # Execute the dbt CLI command in a subprocess.
-    full_command = " ".join(command_list) # make sure all are strings before joining for printing
+    full_command = " ".join(
+        command_list
+    )  # make sure all are strings before joining for printing
```
chamini2, line 83: you can actually remove the comment from THIS line and leave it on the above one.

src/fal/cli/dbt_runner.py
```diff
@@ -43,7 +46,7 @@ def logs(self) -> List[Dict[str, Any]]:
         return self._logs
 
 
-def dbt_run(args, models_list: List[str]):
+def dbt_run(args, models_list: List[str], target_path: str, run_index: int = 0):
```
chamini2, line 49: maybe we should NOT offer a default for it, to avoid bugs of overwriting old files.

src/fal/cli/dbt_runner.py
chamini2: and remove the comment here
```suggestion
    full_command = " ".join(command_list)
```

chamini2 (approved):
little detail (not really related, but your formatter brought it up)

PR #224, opened 2022-04-01
write_to_source overwrite mode

chamini2:


src/faldbt/project.py
```diff
@@ -397,28 +414,30 @@ def write_to_source(
         data: pd.DataFrame,
         target_source_name: str,
         target_table_name: str,
+        # TODO: Make dtype named-param in the future?
         dtype: Any = None,
+        *,
+        mode: Literal["append", "overwrite"] = "append",
```
mkam039, line 420: Ideally this is an enum, no? Or maybe at least encapsulated in string variables.

mkam039 (approved):

PR #229, opened 2022-04-04
Project name should be part of model id

turbo1912:


mkam039 (approved):

tests/graph/test_selectors.py
```diff
@@ -4,18 +4,20 @@
 from argparse import Namespace
 from utils import assert_contains_only
 
+PROJECT_NAME = "fal"
```
chamini2, line 7: maybe something like `test_project` to not confuse this with the cli tool fal?

chamini2 (approved):

PR #230, opened 2022-04-05
Allow models without select flag

turbo1912:


src/fal/cli/dbt_runner.py
```diff
@@ -73,15 +72,26 @@ def dbt_run(args, models_list: List[str], target_path: str, run_index: int):
...
+    # make sure all are strings before joining for printing
+    command_list = list(map(str, command_list))
+
     # Execute the dbt CLI command in a subprocess.
-    full_command = " ".join(
-        command_list
-    )  # make sure all are strings before joining for printing
+    full_command = " ".join(command_list)
+
+    # make sure all are strings before joining for printing
```
chamini2, line 94: remove this comment, it's copied from above and not relevant here.

src/fal/cli/dbt_runner.py
```diff
@@ -73,15 +72,26 @@ def dbt_run(args, models_list: List[str], target_path: str, run_index: int):
     if args.selector:
```
chamini2, line 72: Shouldn't we remove this too? I think a selector is processed in the fal flow level and then we pass the correct `--select` flags to dbt.

turbo1912: yea i am not sure how we can support selectors right now seems like the currently code probably doesn't handle it perfectly either.. for example we dont handle tags.. ok creating a ticket for it

chamini2: I am not sure it doesn't work, I haven't looked too much in detail inside fal flow selection flags process. I think it _should_ work? Do you want to add an integration test to test it?

turbo1912: Adding tests in a different PR..

PR #238, opened 2022-04-07
Add select support for complex selectors like tags and state

turbo1912:


mkam039 (approved):

integration_tests/features/steps/fal_steps.py
```diff
@@ -98,7 +102,7 @@ def invoke_failing_fal_flow(context):
 
 @then("model named {model_name} is removed")
 def delete_model(context, model_name):
-    os.system(f"rm {context.base_dir}/other_models/{model_name}.sql")
+    shutil.rmtree(os.path.join(context.base_dir, "other_models"))
```
chamini2, line 105: I think you are deleting the whole directory instead of `model_name`?

chamini2: What if we add an after_each behave hook that empties the `other_models` directory? We could get rid of this `@then`.

src/fal/cli/selectors.py
```diff
@@ -80,15 +93,45 @@ class SelectorPlan:
...
+
+        if self.type == SelectType.MODEL:
+            self.unique_ids = [f"model.{project_name}.{node_name}"]
+        elif self.type == SelectType.SCRIPT:
             self.unique_ids = _expand_script(node_name, unique_ids)
+        elif self.type == SelectType.COMPLEX:
+            self.unique_ids = unique_ids_from_complex_selector(selector, fal_dbt)
+
+
+def unique_ids_from_complex_selector(select, fal_dbt: FalDbt) -> List[str]:
```
chamini2, line 114: this does not handle stuff like `tag:x+`, right? It would ignore scripts associated to those.

What if we do a recursive solution? we pass the resulting models we got from `tag:x` (let's say models `[a, c, e]`) to `SelectorPlan` again as `a+ c+ e+`.

Meaning: we process colons in general IN dbt and then process the resulting models as normal models we di here.

src/fal/cli/selectors.py
```diff
@@ -68,6 +75,12 @@ def contains_script_name(id: str):
     return list(filter(lambda id: contains_script_name(id), unique_ids))
```
chamini2, line 75: couldn't you just say `list(filter(contains_script_name, unique_ids))`?

src/fal/cli/selectors.py
chamini2: Maybe it does, I am a little bit lost, to be honest.

src/faldbt/project.py
```diff
@@ -175,8 +175,8 @@ def __init__(self, nativeRunResult: Optional[RunResultsArtifact]):
 @dataclass
 class CompileArgs:
     selector_name: Optional[str]
-    select: Tuple[str]
-    models: Tuple[str]
+    select: List[str]
```
chamini2, line 178: I think these come as tuples from the `cli` parsing. Maybe change there too?

integration_tests/features/steps/fal_steps.py
```diff
@@ -60,7 +61,10 @@ def persist_state(context, folder_name):
 
 @when("model named {model_name} is added")
 def add_model(context, model_name):
-    os.system(f"echo 'select 1' > {context.base_dir}/other_models/{model_name}.sql")
+    os.mkdir(os.path.join(context.base_dir, "other_models"))
+    f = open(f"{context.base_dir}/other_models/{model_name}.sql", "a")
+    f.write("select 1")
+    f.close()
```
chamini2, line 67:
```suggestion
with open(f"{context.base_dir}/other_models/{model_name}.sql", "a") as f:
    f.write("select 1")
```

chamini2: Also, why mode `a` and not `w`?

src/fal/cli/selectors.py
turbo1912: it does.. I added tests for it

integration_tests/features/flow_run.feature
```diff
@@ -87,3 +87,25 @@ Feature: `flow run` command
...
+    """
+    Then the following models are calculated:
+    | agent_wait_time |
+
+  Scenario: fal flow run command with complex selectors and children
+    Given the project 001_flow_run_with_selectors
+    When the data is seeded
+    When the following command is invoked:
+    """
+    fal flow run --profiles-dir $profilesDir --project-dir $baseDir --select tag:daily+
```
chamini2, line 106: amazing

src/fal/cli/selectors.py
turbo1912: see the selector logic we parse if it requires parents or children so same thing applies to the complex tags

integration_tests/features/steps/fal_steps.py
chamini2: But you changed the behavior, you are deleting all contents of the directory, not the model file as it was done before this PR

turbo1912: Is this intentional that this folder is committed? 

https://github.com/fal-ai/fal/tree/main/integration_tests/projects/001_flow_run_with_selectors/other_models

chamini2 (approved):
Great, great!

PR #242, opened 2022-04-07
Generate .txt files instead of .py files for integration tests to easily ignore them

chamini2:


integration_tests/.gitignore
```diff
@@ -0,0 +1 @@
+*.txt
```
turbo1912, line 1: are we not generating these in the temp_dir already? why do we need to ignore them?

turbo1912 (approved):

PR #248, opened 2022-04-18
Update docs with diagrams of fal flow

turbo1912:


chamini2 (commented):
add a note somewhere saying it runs the yellow border nodes

PR #252, opened 2022-04-19
Clear the dbt relations cache every time we write to it

chamini2:


turbo1912 (approved):

src/faldbt/lib.py
```diff
@@ -67,7 +67,9 @@ def initialize_dbt_flags(profiles_dir: str):
 def _get_adapter(project_dir: str, profiles_dir: str) -> SQLAdapter:
     config = parse.get_dbt_config(project_dir, profiles_dir)
 
-    return adapters_factory.get_adapter(config)  # type: ignore
+    adapter: SQLAdapter = adapters_factory.get_adapter(config)  # type: ignore
+
+    return adapter
```
turbo1912, line 72: why is this necessary?

chamini2: I had modified the adapter before returning and then removed that but did not remove the "store it in a variable" code I introduced

PR #263, opened 2022-04-22
Add vars flag for fal flow run and pass to dbt 

### LINKED ISSUE #253: Pass command line variables to fal flow
Is there a way to pass command line variables to a fal flow run like we do for a dbt run?

$ dbt run --vars '{"key": "value"}'

chamini2:
Maybe we should set environment variables in the scripts when running. But if we do that, we should read `dbt_project.yml#vars` too.

closes #253

### DIFF
--- a/integration_tests/features/steps/fal_steps.py
+++ b/integration_tests/features/steps/fal_steps.py
@@ -72,14 +72,19 @@ def invoke_command(context):
     context.exc = None
     try:
         cli(shlex.split(args))
-    except Exception as e:
-        context.exc = e
+    except Exception:
+        import sys
+
+        context.exc = sys.exc_info()
 
 
-@then("it throws an {type} exception with message '{msg}'")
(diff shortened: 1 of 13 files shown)


integration_tests/features/steps/fal_steps.py
```diff
@@ -63,21 +63,28 @@ def add_model(context, file):
...
+
+        context.exc = sys.exc_info()
 
 
-@then("it throws an {type} exception with message '{msg}'")
-def invoke_command_error(context, type, msg):
-    assert isinstance(context.exc, eval(type)), "Invalid exception - expected " + type
-    assert msg in str(context.exc), "Invalid message - expected " + msg
+@then("it throws an {etype} exception with message '{msg}'")
+def invoke_command_error(context, etype: str, msg: str):
```
turbo1912, line 82: 👍

turbo1912 (approved):
wow what a complete PR, awesome

PR #266, opened 2022-04-25
Make `--exclude` work with `fal flow run`

mederka:


integration_tests/features/flow_run.feature
```diff
@@ -138,3 +138,30 @@ Feature: `flow run` command
       | zendesk_ticket_data |
     And the script zendesk_ticket_data.check_extra.py output file has the lines:
       | extra_col: yes |
+
+  Scenario: fal flow run command with exclude arg
```
turbo1912, line 142: can you add a test that has both `--select` and `--exclude` at the same time

mederka: Sure

chamini2 (approved):
Good!!

PR #271, opened 2022-04-27
Add Pure Python models doc

chamini2:
This is incomplete, but the idea of what I want to write.


docsite/docs/Docs/fal-cli/pure-python-models.md
```diff
@@ -0,0 +1,129 @@
+---
+sidebar_position: 8
+---
+
+# Making pure Python dbt models
```
mederka, line 5: This section reads more like an intro to a blog post rather than a documentation section. I would only keep line 7 and clarify it.

docsite/docs/Docs/fal-cli/pure-python-models.md
```diff
@@ -0,0 +1,129 @@
...
+In this guide we will build a dbt model with a Python script using fal.
+
+This guide is inspired by Andy Reagan's post where fal is mentioned, [Non-SQL in your DBT pipeline](https://andyreagan.medium.com/non-sql-in-your-dbt-pipeline-c7cef2091619). Which in turn is inspired by Claire Carroll's post on the dbt Discourse, [Representing non-SQL models in a dbt DAG](https://discourse.getdbt.com/t/representing-non-sql-models-in-a-dbt-dag/2083).
+
+<!-- TODO: Why would we do this? -->
+We love dbt, but it only runs SQL, some stuff can only (or much more easily) be achieved in other programming languages. In this case, we are going to use Python.
+
+## Context
+
+We are going to expand the jaffle shop classic example of dbt. We will add a life time value prediction to the shop costumers.
```
mederka, line 16: We need to explain what "life time value prediction" means.

docsite/docs/Docs/fal-cli/pure-python-models.md
```diff
@@ -0,0 +1,129 @@
...
+*/
+
+-- Check Python file for real behaviour
+SELECT * FROM {{ target.schema }}.{{ model.name }}
+```
+> File: `models/customers_predicted_ltvs.sql`
+
+Things to note:
+
+- Including the `ref`s in a comment means that the edges in your DAG will correctly drawn between other models and this non-SQL model
```
mederka, line 43: will be correctly drawn

docsite/docs/Docs/fal-cli/pure-python-models.md
```diff
@@ -0,0 +1,129 @@
...
+- Declaring this as an `ephemeral` model means dbt will not build this model
+- Now this can be documented with `descriptions` in the `schema.yml` file
```
mederka, line 45: Not sure if this is also a bullet point.

docsite/docs/Docs/fal-cli/pure-python-models.md
```diff
@@ -0,0 +1,129 @@
...
+We are going to expand the jaffle shop classic example of dbt. We will add a life time value prediction to the shop costumers.
+
+## Make dbt _know_ about the model
+
+The first thing we want to do is to actually create a dbt model. 
+
+```sql
+{{ config(materialized='ephemeral') }}
+/*
+This table is generated by fal.
```
mederka, line 25: This comment is confusing. What does `this` mean? Which table are we talking about?

Also, do we need to specify the `fal flow` requirement here in this comment? Shouldn't it be somewhere more visible?

docsite/docs/Docs/fal-cli/pure-python-models.md
```diff
@@ -0,0 +1,129 @@
...
+```
+> File: `models/customers_predicted_ltvs.sql`
+
+Things to note:
+
+- Including the `ref`s in a comment means that the edges in your DAG will correctly drawn between other models and this non-SQL model
+- Declaring this as an `ephemeral` model means dbt will not build this model
+- Now this can be documented with `descriptions` in the `schema.yml` file
+
+And let's add it to the schema.
```
mederka, line 47: I don't think we need to show `schema.yml` twice. It's better to show it with the script attached, in it's final state.

docsite/docs/Docs/fal-cli/pure-python-models.md
```diff
@@ -0,0 +1,129 @@
...
+      fal:
+        scripts:
+          after:
+            - models/customers_predicted_ltvs.py
+```
+> File: `models/schema.yml`
+
+## Reference the created model downstream
+
+Now you can reference it from downstream models and it will know what to do.
```
mederka, line 109: Should this model actually do something?

docsite/docs/Docs/fal-cli/pure-python-models.md
```diff
@@ -0,0 +1,129 @@
...
+),
+...
+```
+> File: `models/fct_customers.sql`
+
+<!-- TODO: add a picture of the docs with the node in there -->
+
+## Run the models with fal flow
+
+The last step is to test it. We are using the `--experimental-flow` flag to
```
mederka, line 123: Not sure if "test" is the right word, as it means something different in dbt context.

docsite/docs/Docs/fal-cli/pure-python-models.md
```diff
@@ -0,0 +1,129 @@
...
+...
+```
+> File: `models/fct_customers.sql`
+
+<!-- TODO: add a picture of the docs with the node in there -->
+
+## Run the models with fal flow
+
+The last step is to test it. We are using the `--experimental-flow` flag to 
+run fal after scripts right after a dbt node has finished to before continuing with downstream nodes.
```
mederka, line 124: "right after a dbt node has finished to before continuing " 

couldn't parse this

chamini2:
This is no longer a solution, since we have pure python models.

PR #278, opened 2022-04-29
Add different Python versions testing daily

chamini2:
Profiles tests with version: https://github.com/fal-ai/fal/actions/runs/2260749676

What I just introduced for profile tests creates a branch for mutex control per profile: https://github.com/fal-ai/fal/branches/all?query=gh-mutex

And means we CANNOT CANCEL an integration test workflow/job because there is a mutex post run step.

This is all until they support it in Github : https://github.com/github/feedback/discussions/5435


chamini2:
Link to profiles tests: https://github.com/fal-ai/fal/runs/6233303974?check_suite_focus=true

turbo1912 (approved):

.github/workflows/test_python_versions.yml
```diff
@@ -0,0 +1,61 @@
+name: Run Integration Tests Python Versions
```
mederka, line 1: Instead of having a separate workflow, why don't we add postgres to daily test_profiles?

mederka (approved):
Great! Just a small comment.

.github/workflows/test_python_versions.yml
chamini2: I think that is a good idea. I'll do it

chamini2: We did lose testing different dbt versions against different python versions. But on second thought, I don't think this adds much value, since the dbt code is already tested by dbt-labs/dbt-core repo.

PR #280, opened 2022-04-29
Merge run_results files of multiple dbt runs

turbo1912:


src/fal/cli/flow_runner.py
```diff
@@ -95,3 +108,55 @@ def _unique_id_to_model_name(unique_id: str):
...
+    if exists(target_path):
+        fal_results_regex = re.compile("fal_results_\\d*.json")
+        run_result_files = filter(
+            lambda dir: isfile(os.path.join(target_path, dir))
+            and bool(fal_results_regex.match(dir)),
+            os.listdir(target_path),
+        )
+        return list(map(lambda dir: os.path.join(target_path, dir), run_result_files))
+    else:
+        return []
```
chamini2, line 162:
```suggestion
        run_result_files = filter(lambda p: p.is_file(), Path(target_path).glob("fal_results_*.json"))
        return list(run_result_files)
```

And change return type to `List[Path]` or `map` over the list before returning to with str

```
        return list(map(str, run_result_files))
```

src/fal/cli/flow_runner.py
```diff
@@ -95,3 +108,55 @@ def _unique_id_to_model_name(unique_id: str):
...
+            result.append(item)
+    return result
+
+
+def _get_all_result_content(file) -> Dict:
+    with open(file) as content:
+        return json.load(content)
+
+
+def _get_result_array(file) -> List[str]:
```
chamini2, line 148:
```suggestion
def _get_result_array(file) -> List[dict]:
```

src/fal/cli/flow_runner.py
```diff
@@ -95,3 +108,55 @@ def _unique_id_to_model_name(unique_id: str):
 
 def _unique_ids_to_model_names(id_list: List[str]):
     return list(map(lambda id: _unique_id_to_model_name(id), id_list))
+
+
+def _combine_fal_run_results(target_path):
+    last_run_result_file = os.path.join(target_path, "run_results.json")
+    last_run_result = _get_all_result_content(last_run_result_file)
+    last_run_result_array: List[str] = last_run_result["results"]
```
chamini2, line 116:
```suggestion
    last_run_result_array = _get_result_array(os.path.join(target_path, "run_results.json"))
```

src/fal/cli/flow_runner.py
```diff
@@ -95,3 +108,55 @@ def _unique_id_to_model_name(unique_id: str):
...
+        file.close()
+
+
+def _distinct_by(target_list, key):
+    seen_keys = []
+    result = []
+    for item in target_list:
+        if item[key] not in seen_keys:
+            seen_keys.append(item[key])
+            result.append(item)
```
chamini2, line 139: interesting what we are giving priority to here. We are keeping the OLDEST `fal_results_X` values.

I have two options when `item[key] in seen_keys`

1. We should or keep the NEWEST value
2. We should raise an exception because we should not run the same unique id twice.

chamini2 (changes requested):
I don't know if this already happens in the code, but we should DELETE all the `fal_results_X.json` files at the end of a run. If not, we could have one run interfering with the next one.

src/fal/cli/flow_runner.py
turbo1912: 👍

src/fal/cli/flow_runner.py
turbo1912: 👍

src/fal/cli/flow_runner.py
turbo1912: I need `last_run_result` below in the same function

src/fal/cli/flow_runner.py
turbo1912: as discussed we dont even need this

turbo1912:
> I don't know if this already happens in the code, but we should DELETE all the `fal_results_X.json` files at the end of a run. If not, we could have one run interfering with the next one.

so about this, `fal_results_X.json` or just `fal_result.json`. I think its just `fal_results.json`

chamini2:
> > I don't know if this already happens in the code, but we should DELETE all the `fal_results_X.json` files at the end of a run. If not, we could have one run interfering with the next one.
> 
> so about this, `fal_results_X.json` or just `fal_result.json`. I think its just `fal_results.json`

Once a dbt run is done, you copy results to a `fal_results_X.json` (X is a number) file:
https://github.com/fal-ai/fal/blob/2b80ec4b860cde6e8ee17dd478f030bbbd914e48/src/fal/cli/dbt_runner.py#L146-L151

You then get all the `fal_results_*.json` files and combine them:
https://github.com/fal-ai/fal/blob/2b80ec4b860cde6e8ee17dd478f030bbbd914e48/src/fal/cli/flow_runner.py#L139-L144

So if we had a flow run with 5 dbt runs (generated `fal_results{1,5}.json`) and then the next flow run has 2 dbt runs, it will take results from the old run and combine them as if they had just happened.

turbo1912:
> So if we had a flow run with 5 dbt runs (generated `fal_results{1,5}.json`) and then the next flow run has 2 dbt runs, it will take results from the old run and combine them as if they had just happened.

got it I think both had to happen, run_results.json between dbt runs need to be deleted and run_results_*.json files between `fal flow` runs need to be deleted

src/fal/cli/flow_runner.py
```diff
@@ -87,11 +102,53 @@ def _flow_node_to_fal_scripts(list: List[Union[FalFlowNode, None]]) -> List[FalS
...
+        result_files,
+        [],
+    )
+
+    last_run_result = _get_all_result_content(result_files[-1])
+    last_run_result[RUN_RESULTS_KEY] = all_results
+    run_result_file = os.path.join(target_path, RUN_RESULTS_FILE_NAME)
+
+    # remove all run_results_*.json files for the next fal flow run
+    _remove_all(result_files)
```
chamini2, line 128: nitpicking: this could not run if there was an exception in a user script and would leave the target "dirty".

Also, what `run_results.json` do we create if there is a failure.

Maybe for a different ticket/PR, but to consider.

chamini2 (approved):
Thank you!

PR #290, opened 2022-05-06
WIP: Make write functions work on Athena

mederka:
- Process connection data from profile
- Update rename_relation to handle Athena renaming
- Pass location to adapter dialect properties

### DIFF
--- a/src/faldbt/lib.py
+++ b/src/faldbt/lib.py
@@ -1,18 +1,20 @@
 # NOTE: INSPIRED IN https://github.com/dbt-labs/dbt-core/blob/43edc887f97e359b02b6317a9f91898d3d66652b/core/dbt/lib.py
 import six
+import os
 from datetime import datetime
 from dataclasses import dataclass
 from uuid import uuid4
-from typing import List, Optional, Tuple, Union
+from typing import List, Optional, Tuple
+from urllib.parse import quote_plus
 
 import dbt.version
 import dbt.semver
 import dbt.flags as flags
 import dbt.adapters.factory as adapters_factory
 from dbt.contracts.connection import AdapterResponse
-from dbt.logger import GLOBAL_LOGGER as logger
 from dbt.adapters.sql import SQLAdapter
 from dbt.adapters.base import BaseRelation
+from dbt.contracts.graph.compiled import CompileResultNode
 
 from . import parse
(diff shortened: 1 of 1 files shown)


mederka:
We would need to update integration tests to run athena profile daily as well.

mederka:
requires installation of pyathena==2.5.4 which is in conflict with dbt-athena

mederka:
It works now if you install `dbt-athena` from our branch and update pyathena:

```
pip install git+https://github.com/fal-ai/dbt-athena.git@update-pyathena-to-2.6
```

turbo1912 (approved):

src/faldbt/lib.py
```diff
@@ -341,4 +366,31 @@ def _alchemy_engine(adapter: SQLAdapter, database: Optional[str]):
     def null_dump(sql, *multiparams, **params):
         pass
 
+    if adapter.type() == "athena":
+        AWS_ACCESS_KEY = os.environ["AWS_ACCESS_KEY_ID"]
+        AWS_SECRET_KEY = os.environ["AWS_SECRET_ACCESS_KEY"]
```
chamini2, line 371: Doesn't dbt need these credentials too? Why do we need it from environ?

src/faldbt/lib.py
```diff
@@ -258,6 +257,21 @@ def _write_relation(
...
+            "location": f"{s3_dir}{alchemy_table.schema}/{alchemy_table.name}/",
+            "tblproperties": None,
+            "compression": None,
+            "bucket_count": None,
+            "row_format": None,
+            "serdeproperties": None,
+            "file_format": None,
+            "partition": None,
+            "cluster": None,
+        }
```
chamini2, line 273: instead of setting all these `None`, can you do `alchemy_table.dialect_options["awsathena"]["location"] = ...` or is `alchemy_table.dialect_options["awsathena"]` not set?

mederka: These are checked one by one by PyAthena but without using a getter, so we need to define them.

chamini2: Yes, but I meant that maybe the already exist in the `alchemy_table.dialect_options["awsathena"]` object and just set the specific `location` property.

Issue #301, opened 2022-05-09
Pure Python models design


chamini2:
Introduce Python-only models to fal. This mirrors the dbt models for SQL. A Python model will _be_ a script with a call to `write_to_model` that creates the table in the database. Python models can be referenced from other dbt models as well as python models in the project using the ref function.

- Python models can be used interchangeably with any dbt model in the fal flow graph selectors
- Python models appear in the dbt lineage graph
- Python models can be documented the same way dbt models can in the schema.yml file

*****

Until now we have only attached Python scripts to existing SQL models and have done some creative workarounds to get the data back into the data warehouse. If we want to go closer to what seems to be the dbt ideas on how a transformation layer should work, we need to be able to create a new dbt model with the data from other models instead of modifying existing ones. Doing this enables us to be confident about parallelizations because no 2 scripts could be writing to the same model (race conditions).

# Previous work

Based on the [post from Andy Reagan](https://andyreagan.medium.com/non-sql-in-your-dbt-pipeline-c7cef2091619), we can see that if we generate a dbt model with

```sql
{{ config(materialized='ephemeral') }} 

/*
List your dbt dependencies here:

{{ ref('customers') }}
{{ ref('orders') }}
*/

select * {{ target.schema }}.{{ model.name }}
```

We can make dbt believe that the table is there and be able to reference it from other models ([read more on the ephemeral config](https://docs.getdbt.com/docs/building-a-dbt-project/building-models/materializations#ephemeral) to understand how this hack works), while not really running the code for it.

# Proposal

The usage would be:

1. realize there are some Python scripts that are to be treated as models
  a. understand dependencies the Python script has to other models
2. write the ephemeral file for each of the models (with dependencies as a comment)
2. run the models
2. delete the ephemeral file for each of the models
  a. consider how this would affect `dbt docs`?

Let's assume the user has the following project

```
my_project/
├── dbt_project.yml
├── fal_scripts
│   └── after.py
└── models
    ├── some_model.sql
    ├── my_model.py
    └── schema.yml
```

Where the script looks like this:

> file `models/my_model.py`
```py
df = ref('some_model') # how do we realize there is a dependency?

s = source('app', 'a_source')

# use the df...

write_to_model(df)
```

### Mimic dbt

Leverage [the `model-paths` configuration](https://docs.getdbt.com/reference/project-configs/model-paths) to look for `.py` files and assume these are Python models.

The user would just write a model in one of the model paths.

[...]

PR #306, opened 2022-05-12
Python model generation

chamini2:
~~We have to stop parsing the whole dbt project before generating these and instead do a "lightweight" parsing to just get source-dirs. If not, dbt analyzes the SQL files and realizes that some SQL files do not exist (they will exist after generation).~~ DONE

Fixes FEA-55 and #301


src/fal/cli/dbt_model_generator.py
```diff
@@ -0,0 +1,132 @@
...
+            file.write(sql_contents)
+
+
+def _check_path_safe_to_write(sql_path: Path, py_path: Path):
+    if sql_path.exists():
+        with open(sql_path, "r") as file:
+            contents = file.read()
+            if "FAL_GENERATED" not in contents:
+                raise RuntimeError(
+                    f"File '{sql_path}' not generated by fal would be overwritten by generated model of '{py_path}'. Please delete or move."
```
mederka, line 49: "Please rename or remove" ?

src/fal/cli/dbt_model_generator.py
```diff
@@ -0,0 +1,132 @@
...
+                )
+
+
+def _find_python_files(source_dirs: Union[List[str], List[Path]]) -> List[Path]:
+    return list(
+        flatten(
+            map(
```
mederka, line 56: Indentation here looks strange

src/fal/cli/dbt_model_generator.py
```diff
@@ -0,0 +1,132 @@
...
+                lambda d: (p for p in Path(d).rglob("*.py") if p.is_file()),
```
mederka, line 57: Should define this lambda as a separate function? We will also need to process .ipynb files.

src/fal/cli/dbt_model_generator.py
```diff
@@ -0,0 +1,132 @@
...
+
+__deps__
+
+*/
+
+SELECT * FROM {{ target.schema }}.{{ model.name }}
+"""
+
+
+def generate_python_dbt_models(fal_dbt: FalDbt):
```
mederka, line 26: Why don't we generate python dbt models before initiating `fal_dbt`? All we need is a project dir.

src/fal/cli/flow_runner.py
```diff
@@ -23,6 +24,11 @@
 
 def fal_flow_run(parsed: argparse.Namespace):
     fal_dbt = create_fal_dbt(parsed)
+
+    if parsed.experimental_python_models:
+        telemetry.log_call("experimental_python_models")
+        generate_python_dbt_models(fal_dbt)
```
mederka, line 30: In the end we will have to clean these up, right?

mederka (commented):
Very interesting implementation, @chamini2. I like the use of `ast`.

The biggest comment I have is why do we need to initiate `FalDbt` before generating SQL files? As long as we know the `project_dir` and can read `dbt_project.yml`, we should be able to find all the relevant python files.

src/fal/cli/dbt_model_generator.py
chamini2: lol, it's actually fine, 4 spaces each time

src/fal/cli/dbt_model_generator.py
chamini2: Not sure if we will process ipynb as models or just as hooks.

src/fal/cli/dbt_model_generator.py
```diff
@@ -0,0 +1,127 @@
...
+    '''
+    Simple regex analysis for docstring in top of the file. User can list dependencies one per line, but not multiline.
+    Example:
+
+    """
+    A Python model with some docstring introduction.
+
+    Dependencies:
+    - ref('model')
+    - source('some', 'table')
```
turbo1912, line 95: do we need source dependencies here? Dont we need this just for dependency resolution? Does listing the source actually change anything?

turbo1912 (approved):

src/fal/cli/dbt_model_generator.py
chamini2: You are right, they probably don't do anything. I can just remove source for this.

PR #307, opened 2022-05-13
Automatically handle aws credentials

mederka:


chamini2 (approved):
Awesome, can't believe this was it.

PR #308, opened 2022-05-13
Offer `init_fal` iPython magic command

mederka:
Initialize fal magic variables inside a notebook. Must provide `project_dir` and `profiles_dir`.

In a notebook:
```
from faldbt.magics import init_fal
```

This makes `init_fal` line magic command available:
```
%init_fal project_dir=/my_project_dir profiles_dir=/my_profiles_dir
```

Now fal magic functions are available:
```
my_models = list_models()
my_df = ref('my_model')
```

- Couldn't find a simple way to test this, use of magic decorators requires being inside a notebook runtime.

### DIFF
--- a/src/faldbt/magics.py
+++ b/src/faldbt/magics.py
@@ -1,0 +1,41 @@
+from IPython.core.magic import register_line_magic, needs_local_scope
+from fal import FalDbt
+
+
+@register_line_magic
+@needs_local_scope
+def init_fal(line="", local_ns={}):
+    '''
+    Init fal magic variables. Must provide project_dir and profiles_dir.
+
+    Example:
+    """
+    from faldbt.magics import init_fal
+
+    %init_fal project_dir=/my_project_dir profiles_dir=/my_profiles_dir
+    """
+    '''
+    args = dict([arg.split("=") for arg in line.split()])
+    if not args.get("project_dir") or not args.get("profiles_dir"):
+        raise Exception(
+            """
+            Both project_dir and profiles_dir need to be provided:
+            Example: %init_fal project_dir=/my_project_dir profiles_dir=/my_profiles_dir
+            """
+        )
+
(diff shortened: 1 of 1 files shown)


src/faldbt/magics.py
```diff
@@ -0,0 +1,36 @@
...
+from fal import FalDbt
+
+
+@register_line_magic
+@needs_local_scope
+def init_fal(line="", local_ns={}):
+    """
+    Init fal magic variables. Must provide project_dir and profiles_dir.
+
+    Example: %init_fal project_dir=/my_project_dir profiles_dir=/my_profiles_dir
```
chamini2, line 11: include the `from faldbt.magics import init_fal` to the example?

src/faldbt/magics.py
```diff
@@ -0,0 +1,36 @@
...
+            Example: %init_fal project_dir=/my_project_dir profiles_dir=/my_profiles_dir
+            """
+        )
+
+    faldbt = FalDbt(args["project_dir"], args["profiles_dir"])
+
+    fal_globals = {
+        "ref": faldbt.ref,
+        "source": faldbt.source,
+        "write_to_source": faldbt.write_to_source,
```
chamini2, line 27: I think you are missing `faldbt.write_to_model`. But I wonder if passing `faldbt` as the global is enough?

```py
    local_ns.update(faldbt)
```

Although maybe `local_ns` *needs* a `dict` object.

chamini2 (approved):
This is super cool!!

src/faldbt/magics.py
mederka: Good catch! Also missing `context`.

The idea is that the notebook runtime should be as similar as possible to fal runtime and in notebook runtime we don't have `context` and `write_to_model`. Should we have them? How would we accomplish it? What if one notebook is associated with two models?

PR #320, opened 2022-05-20
fix: Change dbt functions args check for ast.Str that works on all versions

chamini2:
> Deprecated since version 3.8: Old classes ast.Num, ast.Str, ast.Bytes, ast.NameConstant and ast.Ellipsis are still available, but they will be removed in future Python releases. In the meantime, instantiating them will return an instance of a different class.

So we need to check `ast.Str` for 3.7

closes FEA-83

### DIFF
--- a/src/fal/cli/model_generator/deps_generator.py
+++ b/src/fal/cli/model_generator/deps_generator.py
@@ -41,11 +41,26 @@ def _find_dbt_function_calls(calls: List[ast.Call], func_name: str) -> List[ast.
         if isinstance(call.func, ast.Name) and call.func.id == func_name
     ]
 
-    return [
-        call
-        for call in func_calls
-        if all(map(lambda arg: isinstance(arg, ast.Constant), call.args))
-    ]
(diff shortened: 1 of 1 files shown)


chamini2:
fixed now: https://github.com/fal-ai/fal/runs/6530627497?check_suite_focus=true

src/fal/cli/model_generator/deps_generator.py
```diff
@@ -41,11 +41,26 @@ def _find_dbt_function_calls(calls: List[ast.Call], func_name: str) -> List[ast.
...
+
+def _print_node(node: ast.AST):
+    """
+    For temporary usage during debugging.
+    """
+    print(
+        node,
+        *((f, getattr(node, f)) for f in node._fields),
+        *((f, getattr(node, f)) for f in node._attributes),
+    )
```
mederka, line 63: Do we need this committed?

mederka: It's cool

PR #323, opened 2022-05-24
chore: lowercase generated table names in snowflake

mederka:


src/faldbt/lib.py
```diff
@@ -164,6 +164,12 @@ def _build_table_from_parts(
 ):
     from dbt.contracts.relation import Path, RelationType
 
+    # Avoids automatic quoting by Snowflake
+    if adapter.type() == "snowflake":
+        database = database.lower()
+        schema = schema.lower()
+        identifier = identifier.lower()
+
```
chamini2, line 172: Can we do this in the `_write_relation` function instead? I think this applies to any table to create with sqlalchemy, not the temporal one only

chamini2 (changes requested):

src/faldbt/lib.py
chamini2: I think we should also change the temporal table to have the `__f__` in lowercase. To remove the "hack" we had introduced (it does not really make a difference, but at least tests it for snowflake, right?)

chamini2 (approved):
Thank you!

turbo1912 (approved):

PR #325, opened 2022-05-24
chore: Reusable action for installing dbt and local build of fal

chamini2:
closes FEA-91


chamini2:
one passing is enough to know it works https://github.com/fal-ai/fal/actions/runs/2379455040

PR #333, opened 2022-05-26
fix: Fix BigQuery integration tests

chamini2:
closes FEA-19


chamini2:
image
https://github.com/fal-ai/fal/actions/runs/2392301399

src/faldbt/lib.py
```diff
@@ -117,6 +118,10 @@ def _get_target_relation(
     )
     manifest = parse.get_dbt_manifest(config)
 
+    if adapter.type() == "bigquery":
+        # After creating a table, BigQuery takes some time to realize it is there
+        sleep(2)
+
```
chamini2, line 124: So this is the real _code_ change, all others are test changes

chamini2: Seems like it. I think 1 was enough too, but just to make sure. I say this from the observations I had, not documentation.

PR #340, opened 2022-05-30
feat: `execute_sql` magic function

mederka:
Provides `execute_sql` magic function. Returns a pandas dataframe.

```
my_df = execute_sql("select * from {{ ref('some_model') }}")
```

One caveat: only dbt > v1 is supported. Making it available for v0 would require either a completely different implementation or copying of large chunks of dbt v1 into `faldbt.cp`.

closes FEA-110


.github/workflows/integration_tests.yml
```diff
@@ -47,4 +47,10 @@ jobs:
         working-directory: integration_tests
         env:
           FAL_STATS_ENABLED: false
-        run: behave
+        run: |
+          if [[ '${{ matrix.dbt }}' =~ ^0.*$ ]]
+          then
+            behave --tags ~@dbtv1
```
chamini2, line 53: The way we had said before "exclude these" with a tag was 

```
behave --tags=-broken_profile
```

https://behave.readthedocs.io/en/stable/tutorial.html#controlling-things-with-tags
>then running behave --tags=slow will run just the scenarios tagged @slow. If you wish to check everything except the slow ones then you may run behave --tags=-slow.

integration_tests/features/execute_sql_function.feature
```diff
@@ -0,0 +1,16 @@
+@dbtv1
+Feature: `execute_sql` function
+  Background: Project Setup
+    Given the project 009_execute_sql_function
```
chamini2, line 4: we have a functions project, do we need a new one because it is v1 only?

integration_tests/projects/009_execute_sql_function/fal_scripts/query_other_model.py
```diff
@@ -0,0 +1,20 @@
+import pandas as pd
+import io
+import os
+from functools import reduce
+
+df: pd.DataFrame = execute_sql('SELECT * FROM {{ ref("other_model")}}')
```
chamini2, line 6: This is cool! The `ref` does not get picked up by fal for Python models, though.

integration_tests/projects/009_execute_sql_function/models/other_model.sql
```diff
@@ -0,0 +1,12 @@
+{{ config(materialized='table') }}
+
+WITH data AS (
+
+    SELECT
+        'some text' AS my_text,
+        -- The following column will be filled in an after script with fal
```
chamini2, line 7: We don't actually do this

integration_tests/projects/009_execute_sql_function/models/other_model.sql
```diff
@@ -0,0 +1,12 @@
+{{ config(materialized='table') }}
```
chamini2, line 1: rename the file to not match the other project's model name

integration_tests/projects/009_execute_sql_function/models/some_model.sql
```diff
@@ -0,0 +1,13 @@
+{{ config(materialized='table') }}
```
chamini2, line 1: rename the file to not match the other project's model name

integration_tests/projects/009_execute_sql_function/models/some_model.sql
```diff
@@ -0,0 +1,13 @@
...
+
+WITH data AS (
+    SELECT 
+        cast(1 AS integer) AS my_int,
+        my_text, 
+        -- the after script value should reflect here
```
chamini2, line 7: this is not true anymore

src/faldbt/lib.py
```diff
@@ -67,20 +69,28 @@ def initialize_dbt_flags(profiles_dir: str):
 
 # NOTE: Once we get an adapter, we must call `connection_for` or `connection_named` to use it
 def _get_adapter(
-    project_dir: str, profiles_dir: str, profile_target: str
+    project_dir: str,
+    profiles_dir: str,
+    profile_target: str,
+    config: RuntimeConfig = None,
```
chamini2, line 75: why do we need this?

src/faldbt/project.py
```diff
@@ -620,6 +620,31 @@ def _lazy_setup_firestore(self):
...
 
+    @telemetry.log_call("execute_sql")
+    def execute_sql(self, sql: str) -> pd.DataFrame:
+        """Execute a sql query."""
+        if lib.IS_DBT_V0:
+            raise NotImplementedError(
+                "execute_sql only supported in dbt version >= 1.0.0"
+            )
+
+        from dbt.lib import compile_sql
```
chamini2, line 631: our `faldbt/lib.py` file is actually inspired from `dbt.lib`. Do we want to import it from them or re-do it on our side?

integration_tests/features/execute_sql_function.feature
mederka: Yes. I thought about making it a part of functions project, but in the end this seemed like the tidiest solution, as we we don't have to worry about model selection in other features.

integration_tests/projects/009_execute_sql_function/fal_scripts/query_other_model.py
mederka: That's true! Do we want it to be picked up automatically? I think it would be a great feature.

src/faldbt/project.py
mederka: Great suggestion! But I think that it's better to let dbt handle the `compile_sql` function so we don't have to worry about it.

src/faldbt/lib.py
mederka: A GREAT QUESTION! Short answer: I don't know 🚩🚩🚩

I don't need it when running tests locally but for some reason, running things in gh actions is different. 

Check this log: https://github.com/fal-ai/fal/runs/6679158525?check_suite_focus=true

For some reason it cannot find the profile dir anymore and this happens only after `compile_sql` is called and I only see it in an action worker.

chamini2 (approved):
Add a HACK for the config thing and we can merge!

PR #341, opened 2022-05-31
Add fal-dagster module

mederka:
Lives inside python_modules subdirectory

Is it's own package

We should implement separate workflows for testing and publishing this


mederka:
I think we can shelf this for now

PR #342, opened 2022-05-31
Add duck db support and integration tests

turbo1912:


tests/mock/mockProfile/profiles.yml
```diff
@@ -29,3 +29,9 @@ fal_test:
       dbname: test
       schema: dbt_fal
       threads: 4
+
+duckdb_dbt_project:
+  outputs:
+    default:
+      type: duckdb
+      path: /Users/gorkemyurtseven/dev/duck_db_database/db.duckdb
```
mederka, line 37: Do we need `db.duckdb` committed?

Also, need to fix the path.

tests/test_cli.py
```diff
@@ -7,7 +7,8 @@
 import re
 
 profiles_dir = os.path.join(Path.cwd(), "tests/mock/mockProfile")
-project_dir = os.path.join(Path.cwd(), "tests/mock")
+# project_dir = os.path.join(Path.cwd(), "tests/mock")
+project_dir = "/Users/gorkemyurtseven/dev/duck_db_project/duckdb_dbt_project"
```
mederka, line 11: 🤨

mederka (changes requested):
So it works almost "out-of-the-box"? Awesome!

tests/test_cli.py
turbo1912: i was just running the tests sorry not ready for review!

src/faldbt/lib.py
```diff
@@ -375,6 +375,10 @@ def _alchemy_engine(adapter: SQLAdapter, database: Optional[str]):
         url_string = f"bigquery://{database}"
     if adapter.type() == "postgres":
         url_string = "postgresql://"
+
+    if adapter.type() == "duckdb":
+        url_string = "duckdb://"
+
```
chamini2, line 381: this is not necessary, by default we do 
```
url_string = f"{adapter.type()}://"
```

.github/workflows/test_profiles.yml
```diff
@@ -33,6 +33,11 @@ jobs:
         working-directory: integration_tests
         run: pip install behave
 
+      - name: Install duckdb
+        working-directory: integration_tests
+        if: matrix.profile == 'duckdb'
+        run: pip install duckdb
```
chamini2, line 39: if fal installs `duckdb-engine`, do we need this?

chamini2 (changes requested):

.github/workflows/test_profiles.yml
```diff
@@ -65,6 +70,8 @@ jobs:
           SF_DATABASE: ${{ secrets.SF_DATABASE }}
           SF_WAREHOUSE: ${{ secrets.SF_WAREHOUSE }}
           SF_SCHEMA: ${{ secrets.SF_SCHEMA }}
+          # Duckdb
+          DB_PATH: ${{ github.workspace }}/temp/duck.db
```
chamini2, line 74: if we remove the `/temp` part we can remove the `mkdir temp` step right? Can we do that?

chamini2 (approved):
Great! Just one idea to consider

.github/workflows/test_profiles.yml
turbo1912: yea simpler

PR #346, opened 2022-06-02
tests: Add prefix to alias of integration test projects

chamini2:
One of the BigQuery integration test problems was that we had more than one project using the same table names. This meant that fal tried to replace a view with a table or vice-versa and BigQuery would reject it.

With this new `generate_alias_name` macro for all projects, all tables will be prefixed with the unique identifier. Making them not overlap with one another.

*****

Maybe we should use https://docs.getdbt.com/reference/dbt-jinja-functions/project_name instead of a var `prefix`?

<details>
closes FEA-125
</details>


chamini2:
Now tables are like:

```
1 of 5 START view model dbt_fal.001_s_model_c .................................. [RUN]
2 of 5 START table model dbt_fal.001_zendesk_ticket_data ....................... [RUN]
2 of 5 OK created table model dbt_fal.001_zendesk_ticket_data .................. [SELECT 10 in 0.14s]
1 of 5 OK created view model dbt_fal.001_s_model_c ............................. [CREATE VIEW in 0.14s]
3 of 5 START view model dbt_fal.001_intermediate_model_1 ....................... [RUN]
3 of 5 OK created view model dbt_fal.001_intermediate_model_1 .................. [CREATE VIEW in 0.04s]
4 of 5 START view model dbt_fal.001_intermediate_model_2 ....................... [RUN]
4 of 5 OK created view model dbt_fal.001_intermediate_model_2 .................. [CREATE VIEW in 0.04s]
5 of 5 START view model dbt_fal.001_intermediate_model_3 ....................... [RUN]
5 of 5 OK created view model dbt_fal.001_intermediate_model_3 .................. [CREATE VIEW in 0.05s]
Finished running 1 table model, 4 view models in 0.46s.
```

chamini2:
> Maybe we should use https://docs.getdbt.com/reference/dbt-jinja-functions/project_name instead of a var `prefix`?

I did it. 

Also, what do you guys think of the whole symlink to `default` project?

PR #348, opened 2022-06-03
feat: make source test data available in context

### LINKED ISSUE #336: Support source status and test results in run context
**Context**
fal provides information about model transformation status and their tests, but not for source tests. That prevents a use case of, for example, notifying all the failures in a `dbt build` command.

Ability to access source test results in the `context` of a `fal run`.

Directly parsing manifest files (not a good option considering all `fal` can do).

mederka:
`list_sources` now returns a list of `DbtSource` objects that have a `tests` property and a `status` property.

Closes #336 

closes fea-109

### DIFF
--- a/integration_tests/projects/005_functions_and_variables/fal_scripts/write_to_source_twice.py
+++ b/integration_tests/projects/005_functions_and_variables/fal_scripts/write_to_source_twice.py
@@ -19,6 +19,20 @@ write_to_source(df, "results", "some_source", mode="append")
 source_size = len(source("results", "some_source"))
 output += f"source size {source_size}\n"
 
+for source in list_sources():
+    output += (
+        f"source {source.name} has {len(source.tests)} tests,"
+        f" source status is {source.status},"
+        f" source has {len(source.tests)} tests\n"
+    )
+
+for model in list_models():
+    output += (
+        f"model {model.name} has {len(model.tests)} tests,"
+        f" model status is {model.status},"
+        f" model has {len(model.tests)} tests\n"
+    )
+
 path = reduce(
     os.path.join, [os.environ["temp_dir"], model_name + ".write_to_source_twice.txt"]
 )
--- a/integration_tests/projects/005_functions_and_variables/fal_scripts/write_to_source_twice.py
+++ b/integration_tests/projects/005_functions_and_variables/fal_scripts/write_to_source_twice.py
@@ -22,15 +22,13 @@ output += f"source size {source_size}\n"
 for source in list_sources():
     output += (
         f"source {source.name} has {len(source.tests)} tests,"
-        f" source status is {source.status},"
-        f" source has {len(source.tests)} tests\n"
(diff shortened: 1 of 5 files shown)


integration_tests/projects/005_functions_and_variables/fal_scripts/write_to_source_twice.py
```diff
@@ -19,6 +19,20 @@
...
+        f"source {source.name} has {len(source.tests)} tests,"
+        f" source status is {source.status},"
+        f" source has {len(source.tests)} tests\n"
+    )
+
+for model in list_models():
+    output += (
+        f"model {model.name} has {len(model.tests)} tests,"
+        f" model status is {model.status},"
+        f" model has {len(model.tests)} tests\n"
```
chamini2, line 33: `has {len(model.tests)} tests` happens twice?

src/faldbt/project.py
```diff
@@ -66,16 +73,22 @@ def __post_init__(self):
...
 @dataclass
-class DbtModel:
+class DbtModel(_DbtNode):
     node: ParsedModelNode
-    name: str = field(init=False)
     meta: Dict[str, Any] = field(init=False)
-    status: NodeStatus = field(init=False)
     columns: Dict[str, Any] = field(init=False)
     tests: List[DbtTest] = field(init=False)
     python_model: Optional[Path] = field(init=False)
```
chamini2, line 94: can you add `unique_id` here?

src/faldbt/project.py
```diff
@@ -634,32 +640,40 @@ def _firestore_dict_to_document(data: Dict, key_column: str):
...
     run_results: DbtRunResult, manifest: DbtManifest, generated_models: Dict[str, Path]
 ):
     models = manifest.get_models()
     tests = manifest.get_tests()
+    sources = [
+        DbtSource(
+            name=source.source_name, table_name=source.name, unique_id=source.unique_id
+        )
+        for source in manifest.get_sources()
+    ]
```
chamini2, line 653: What about receiving a source object directly instead of only the parts we need? 
```suggestion
    sources = [DbtSource(source) for source in manifest.get_sources()]
```

chamini2 (changes requested):
We need tests for this

src/faldbt/project.py
```diff
@@ -42,12 +42,19 @@ class FalGeneralException(Exception):
 
 
 @dataclass
-class DbtTest:
-    node: Any
+class _DbtNode:
```
burkaygur, line 45: Is class names with _ common practice?

mederka: i've seen a couple times. It's a way to communicate that a class is not to be imported anywhere else.

src/faldbt/project.py
mederka: We will do a separate refactor where we will re-evaluate use of dataclasses in this file

chamini2 (approved):

PR #352, opened 2022-06-07
refactor: remove unnecessary use of dataclasses in project.py

mederka:


isidentical:
Just for catching up on this: is there a particular reason to move away from dataclasses?

src/faldbt/project.py
```diff
@@ -74,27 +71,30 @@ def __post_init__(self):
...
-    unique_id: str = field(init=False)
-
-    def __post_init__(self):
+    """Holds dbt node information."""
+
+    def __init__(self, node: ParsedModelNode):
+        """Initialize DbtModel with name, status and node."""
+        self.node = node
+
         node = self.node
```
chamini2, line 98: you can remove this line now.

src/faldbt/project.py
```diff
@@ -74,27 +71,30 @@ def __post_init__(self):
...
 
         self.name = node.name
```
chamini2, line 100: What about using

```py
@property
def name(self):
  return self.node.name
```
for these kind of properties?

src/faldbt/project.py
```diff
@@ -41,23 +40,21 @@ class FalGeneralException(Exception):
...
-@dataclass
 class DbtTest(_DbtNode):
-    node: Any
-    model: str = field(init=False)
-    column: str = field(init=False)
+    """Holds dbt test information."""
 
-    def __post_init__(self):
-        node = self.node
+    def __init__(self, node: Any):
```
chamini2, line 55: Shouldn't we be doing?
```py
super().__init__(node.name, node.status)
```

src/faldbt/project.py
```diff
@@ -74,27 +71,30 @@ def __post_init__(self):
...
-    unique_id: str = field()
-    tests: List[DbtTest] = field(init=False)
+    """Holds dbt source information."""
+
+    def __init__(
+        self,
+        name: str,
+        table_name: str,
+        unique_id: str,
+    ):
```
chamini2, line 82: In `DbtTest` we are receiving a `node` arg and then getting all the information needed from it. Should we have the same approach in both classes and receive a `node` here?

Maybe we should consider the same for `_DbtNode`

src/faldbt/project.py
```diff
@@ -163,9 +163,9 @@ def get_sources(self) -> List[ParsedSourceDefinition]:
         return list(self.nativeManifest.sources.values())
 
 
-@dataclass(init=False)
 class DbtRunResult:
-    nativeRunResult: Optional[RunResultsArtifact]
+    """Holds dbt run result information."""
+
     results: Sequence[RunResultOutput]
```
chamini2, line 169: Should we remove this one too?

src/faldbt/project.py
```diff
@@ -163,9 +163,9 @@ def get_sources(self) -> List[ParsedSourceDefinition]:
         return list(self.nativeManifest.sources.values())
 
 
-@dataclass(init=False)
 class DbtRunResult:
-    nativeRunResult: Optional[RunResultsArtifact]
```
chamini2, line 168: Are we losing autocomplete by removing these?

src/faldbt/project.py
isidentical: I think that would be a nice solution as well (if it is viable). If we receive the `node` in all `_DbtNode` subclasses, then we can safely use a dataclass in the `_DbtNode` that takes a single argument (and extracts name etc.) and in each subclass we can do the additional computations in the `__post_init__` (and take additional arguments now, since they wont conflict). This way we don't loose any benefits of the dataclasses (e.g., the auto-generated repr & auto-completion support for various editors) and at the same time keep the code tidier. Though not sure there are any other things that might block this.

mederka: I like @isidentical's suggestion. See updated code.

src/faldbt/project.py
mederka: It's back to being a dataclass, because of the auto-generated __repr__ function.

src/faldbt/project.py
mederka: Not in my editor, but this depends on which language server you're using. Should keep an eye on it.

src/faldbt/project.py
```diff
@@ -71,46 +76,54 @@ def __post_init__(self):
...
-    tests: List[DbtTest] = field(init=False)
+    tests: List[DbtTest] = field(init=False, default_factory=list)
+    status: str = field(init=False, default=None)
 
-    def __post_init__(self):
-        self.tests = []
+    def __repr__(self):
+        attrs = ["name", "tests", "status"]
+        props = ", ".join([f"{item}={repr(getattr(self, item))}" for item in attrs])
+        return f"DbtSource({props})"
```
isidentical, line 90: I think this will be auto-generated, even with `init=False` if I am not missing something.

chamini2: It does not work for `@property` functions

src/faldbt/project.py
```diff
@@ -71,46 +76,54 @@ def __post_init__(self):
...
+                logger.warn(f"Non-generic test was not processed: {node.name}")
 
 
 @dataclass
 class DbtSource(_DbtNode):
-    name: str = field()
-    table_name: str = field()
-    unique_id: str = field()
-    tests: List[DbtTest] = field(init=False)
+    tests: List[DbtTest] = field(init=False, default_factory=list)
```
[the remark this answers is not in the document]
chamini2: We have a weird initialization that happens outside of the class: https://github.com/fal-ai/fal/blob/main/src/faldbt/project.py#L701

This should be improved, but we thought it was outside the scope of this PR.

src/faldbt/project.py
isidentical: Makes sense, didn't think about the `property`.

isidentical (approved):
Amazing work! (thanks @chamini2 for the explanations as well)

src/faldbt/project.py
```diff
@@ -71,46 +76,54 @@ def __post_init__(self):
...
 @dataclass
 class DbtModel(_DbtNode):
-    node: ParsedModelNode
-    meta: Dict[str, Any] = field(init=False)
-    columns: Dict[str, Any] = field(init=False)
-    tests: List[DbtTest] = field(init=False)
-    python_model: Optional[Path] = field(init=False)
-    unique_id: str = field(init=False)
+    tests: List[DbtTest] = field(init=False, default_factory=list)
+    python_model: Optional[Path] = field(init=False, default=None, repr=False)
```
chamini2, line 104: you can remove the `repr` since we define `__repr__`.

mederka:
Thank you, @chamini2 and @isidentical, for your help here 💪 

I will be merging this once the tests succeed.

PR #353, opened 2022-06-07
feat: add support for using the set intersection operator on selectors

isidentical:
Implements the [set intersection operator](https://docs.getdbt.com/reference/node-selection/set-operators#intersections) for selectors.

### DIFF
--- a/integration_tests/features/flow_run_middle_nodes.feature
+++ b/integration_tests/features/flow_run_middle_nodes.feature
@@ -41,3 +41,42 @@ Feature: `flow run` command with py nodes in the middle
       | middle_2.middle_script.py |
     And the script middle_2.middle_script.py output file has the lines:
       | Status: skipped |
+
+  Scenario: fal flow run command with set intersection selector
+    Given the project 008_pure_python_models
+    When the data is seeded
+
+    When the following command is invoked:
+      """
+      fal flow run --profiles-dir $profilesDir --project-dir $baseDir --experimental-models
+      """
+    And the following command is invoked:
+      """
+      fal flow run --profiles-dir $profilesDir --project-dir $baseDir --select model_a+,model_c+ --experimental-models
+      """
+    Then the following models are calculated:
+      | model_d |
+    And the following scripts are ran:
+      | model_c.after.py | model_e.after.py |
+
+  Scenario: fal flow run command with triple intersection selectors on parents
+    When the following command is invoked:
+      """
+      fal flow run --profiles-dir $profilesDir --project-dir $baseDir --select +customers,+orders_forecast,+orders
+      """
+    Then the following models are calculated:
+      | stg_orders | stg_payments |
+
+  Scenario: fal flow run command with intersection and union mixed
+    When the following command is invoked:
+      """
+      fal flow run --profiles-dir $profilesDir --project-dir $baseDir --select +customers +middle_script.py
+      """
+    And the following command is invoked:
+      """
+      fal flow run --profiles-dir $profilesDir --project-dir $baseDir --select stg_customers+,stg_orders+,stg_payments+ middle_script.py
+      """
+    Then the following models are calculated:
+      | customers |
+    Then the following scripts are ran:
+      | middle_2.middle_script.py |
--- a/src/fal/cli/selectors.py
+++ b/src/fal/cli/selectors.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass
 import re
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Iterator
 from fal.node_graph import NodeGraph
 from faldbt.project import CompileArgs, FalDbt
 from dbt.task.compile import CompileTask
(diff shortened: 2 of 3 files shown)


src/fal/cli/selectors.py
```diff
@@ -234,6 +244,8 @@ def depth(self, selector: str) -> Optional[int]:
 OP_PARENTS = SelectorGraphOpDepth(re.compile("^(?P<depth>\\d*)\\+(?P<rest>.*)"))
 OP_CHILDREN = SelectorGraphOpDepth(re.compile("(?P<rest>.*)\\+(?P<depth>\\d*)$"))
 
+# Logic based set operators
+SET_INTERSECTION = SetOperator(",")
```
mederka, line 248: Might be a silly question, but why do we need a class if we only instantiate it once?

integration_tests/features/flow_run_middle_nodes.feature
```diff
@@ -41,3 +41,20 @@ Feature: `flow run` command with py nodes in the middle
...
+    Given the project 008_pure_python_models
+    When the data is seeded
+
+    When the following command is invoked:
+      """
+      fal flow run --profiles-dir $profilesDir --project-dir $baseDir --experimental-models
+      """
+    And the following command is invoked:
+      """
+      fal flow run --profiles-dir $profilesDir --project-dir $baseDir --select model_a+,model_c+ --experimental-models
```
mederka, line 55: Can we have a more explicit example of an intersection? Say we had dag like this:

```
          model_c
        /
model_a
        \
	  model_d
	/
model_b
        \
  	  model_e
```
in which case `fal flow run --select model_a+,model_b+` would only run model d and not c and e.

mederka (commented):
Congrats on your first PR @isidentical !

The implementation is very clear and tidy 👏 

I just have a comment about tests and a small question. See below.

src/fal/cli/selectors.py
isidentical: Right, I should have cleared it a bit more. Currently `fal` does not support using a space-delimited *single* string as an argument to `--select` (it interprets the whole thing [up to the `+`] as one):

```
$ temp_dir=/tmp/fal fal flow run --profiles-dir ../ --experimental-models --select "model_a+ model_c+"
...
networkx.exception.NetworkXError: The node model.fal_008.model_a+ model_c is not in the graph.
```

This is supported by `dbt run`, so I left an open door to extend this operation, and we could probably add a `SET_UNION` in the future with using this class to represent what each separator token (space => `set.union`, comma => `set.intersection`) would correspond to.

```
SET_UNION = SetOperator(" ", action=set.union)
SET_INTERSECTION = SetOperator(",", action=set.intersection)
```

Though I can simply do it in the simple way now, and change in the following PRs (if we decide to fix this bug). WDYT?

integration_tests/features/flow_run_middle_nodes.feature
isidentical: I'll look for a similiar DAG from the existing projects, and add a test for that one too. Thanks!

src/fal/cli/selectors.py
mederka: Thanks for the clarification. I don't have a strong opinion either way, so, my vote would be to keep the class definition.

src/fal/cli/selectors.py
```diff
@@ -66,36 +66,16 @@ def _filter_node_ids(
...
-                output.extend(ids)
+    for selector in selected_nodes:
+        selector_plans = map(
+            lambda sub_selector: SelectorPlan(sub_selector, unique_ids, fal_dbt),
+            SET_INTERSECTION.separate(selector),
+        )
+        selector_outputs = map(
+            lambda plan: set(plan.execute(nodeGraph)), selector_plans
+        )
+        output.extend(set.intersection(*selector_outputs))
```
chamini2, line 78: This looks way cleaner 👏

src/fal/cli/selectors.py
chamini2: So `argparse` does the `SetOperator(" ")` for us and we define `SetOperator(",")`. But how would you do the `action=...` behavior if we don't really have control on applying `SetOperator(" ")`. Would you `join` them?

chamini2: So `argparse` does the `SetOperator(" ")` for us and we define `SetOperator(",")`. But how would you do the `action=...` behavior if we don't really have control on applying `SetOperator(" ")`. Would you `join` them?

isidentical: > So argparse does the SetOperator(" ") for us and we define SetOperator(","). But how would you do the action=... behavior if we don't really have control on applying SetOperator(" "). Would you join them?

We have 2 different cases:
```bash
$ fal flow run --select foo bar
$ fal flow run --select "monty python" eggs "benedict,spam" 
```

Both of them are currently handled by the argparse, and they are parsed like this (by argparse):

```py
["foo", "bar"]
['"monty python", "eggs", '"benedict,spam"']
```

The first case looks perfectly fine, but on the second case (focusing on the initial result, `"monty python"`) we need a second round of splitting. We probably have 2 options:

A) (As you said) Join them all (since the actual argparse splitting behavior is necessary to capture all arguments after select) and split it afters
B) Iterate them, and split each individual item on that list once more. If there are no more spaces in them (like `eggs`), we'll do nothing. But if there are spaces in individual items, we'll split them by space and process them as usual (check for comma etc.)

The first method seems to be a bit complicated, due to the precedence problems (like which thing belongs to which group might be confusing). Though take all of these with a grain of salt, I haven't take a stab at the implementation of this problem yet. But in either way, I think the approach of using `SetOperator` works perfectly fine.

src/fal/cli/selectors.py
isidentical: Thanks!

mederka (approved):
Awesome! Thanks for looking into the tests, @isidentical!

turbo1912:
> Congrats on your first PR @isidentical !

First PR 🎉 😇

integration_tests/features/flow_run_middle_nodes.feature
```diff
@@ -41,3 +41,42 @@ Feature: `flow run` command with py nodes in the middle
       | middle_2.middle_script.py |
     And the script middle_2.middle_script.py output file has the lines:
       | Status: skipped |
+
+  Scenario: fal flow run command with set intersection selector
+    Given the project 008_pure_python_models
+    When the data is seeded
```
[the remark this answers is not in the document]
isidentical: This feature seeds `002_jaffle_store` as its background, but this test in particular uses `008_pure_python_models` so I have to change the context manually on this.

chamini2 (approved):
This is great!

turbo1912 (approved):

PR #360, opened 2022-06-08
tests: add support for testing pure-python models

isidentical:
As discussed in FEA-37, the convention is `<model>.txt` for output files and `<model>.{py, ipynb}` in the feature files. Blocked by: #361

### DIFF
--- a/integration_tests/features/steps/fal_steps.py
+++ b/integration_tests/features/steps/fal_steps.py
@@ -1,5 +1,6 @@
 from functools import reduce
 import os
+from pyexpat import model
 from typing import Dict
 from behave import *
 from fal.cli import cli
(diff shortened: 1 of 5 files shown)


integration_tests/features/steps/fal_steps.py
```diff
@@ -215,19 +233,18 @@ def _get_fal_results_file_name(context):
...
-def _flatten_list(target_list):
+def _flatten(target_iterable):
     flat_list = []
-    for element in target_list:
-        if type(element) is list:
-            for item in element:
-                flat_list.append(item)
+    for element in target_iterable:
+        if isinstance(element, list):
+            flat_list.extend(element)
```
chamini2, line 240: The name change suggests that we could get non-lists, right? (e.g. `_flatten( [(1,2),(3,4)] )`) Not sure if this should give `[(1,2),(3,4)]` or `[1,2,3,4]`.

chamini2 (approved):
This is great!! Thank you.

integration_tests/features/steps/fal_steps.py
```diff
@@ -1,5 +1,6 @@
 from functools import reduce
 import os
+from pyexpat import model
```
turbo1912, line 3: not intentional?

integration_tests/features/steps/fal_steps.py
isidentical: > The name change suggests that we could get non-lists, right?

Yep! For example in `_get_models_from_result`, we pass it a `map()` object instead of a list.

> (e.g. _flatten( [(1,2),(3,4)] )) Not sure if this should give [(1,2),(3,4)] or [1,2,3,4].

I think what flatten is intended right now is just unpacking top-level lists in any iterable (a `map()` object, a list, etc.) to a list. Maybe we could rename it to something like `_unpack_top_level_lists(...)` but not sure. Or we could advance it with adding support to what you said but since this is a test utility, I don't think it would make that much difference. WDYT?

turbo1912: i think its totally fine leave as is..

isidentical:
Rebased, we should be good to go on this!

PR #362, opened 2022-06-08
feat: support quote-grouped `--select` arguments

isidentical:
- Add support for `--select` with `"x y z"` (more discussion [here](https://github.com/fal-ai/fal/pull/353#discussion_r891072798)).
- Remove the `SetOperator` (noticed it was easier if we don't depend on it since we only have 2 operators).

### DIFF
--- a/src/fal/cli/selectors.py
+++ b/src/fal/cli/selectors.py
@@ -1,5 +1,6 @@
-from dataclasses import dataclass
+import itertools
 import re
+from dataclasses import dataclass
 from typing import List, Optional, Union, Iterator
 from fal.node_graph import NodeGraph
 from faldbt.project import CompileArgs, FalDbt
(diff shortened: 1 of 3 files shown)


src/fal/cli/selectors.py
```diff
@@ -61,21 +61,32 @@ def create_plan_from_graph(cls, parsed, nodeGraph: NodeGraph, fal_dbt: FalDbt):
...
+                *[
+                    set(plan.execute(nodeGraph))
+                    for plan in selector_plan_group
+                    # Initially group each set by intersection, and then in
+                    # the next step we'll take a union of all these intersections.
+                ]
+            )
+            for selector_plan_group in selector_plan_groups
+        ]
+        output.extend(set.union(*selector_outputs))
```
chamini2, line 89: Can we add a test like this:

```
# .../integration_tests/projects/008_pure_python_models
❯ dbt ls -s "model_a model_b" , "model_b model_c"
fal_008.model_a
fal_008.model_b
fal_008.fal.staging.model_c
```

I have my doubts because we are processing each `selector` separately. Notice there is a space around `,`

isidentical: Oh, this is an interesting case. I would thought this to be an explicit syntax error, if we'd assume " " as an operator. But apparently it is supported by DBT. I'll check it out, thanks!

chamini2: I think this can be handled by doing `" ".join(select)` and just passing the single string to the selectors code

mederka: As a side-note, I think this is the sort of thing that might change in future versions of dbt and we will have to change with it. How should we keep track of this?

isidentical: Maybe a better way would be using the same algorithm from DBT? As far as I understand, we can't explicitly depend on DBT (since `fal` does not depend on `dbt-core`) but we can still carry over the whole function (it seems pretty-isolated):
https://github.com/dbt-labs/dbt-core/blob/32e1924c3bd67e2e3f8d4bc0ec336682e910cd80/core/dbt/graph/cli.py#L27-L53

We just need to change the representation of `SelectionCriteria` (and stuff like it) to more basic data types.

mederka: Oh yes, that `parse_union` function looks pretty good.

isidentical: True, if we don't depend on dbt-core directly we need to still keep it up to date manually. Is there any links to the upcoming changes that you know? Do they plan to make some backwards incompatible changes on the parsing of unions/intersections?

mederka: Not as far as I know. This is just me being a little paranoid.

chamini2: We can use `from dbt.grapg.cli import parse_union`, it will import the correct one based on the installed dbt version. dbt-core is a peer-dependency for us, I think we don't state it anywhere though.

turbo1912 (approved):

src/fal/cli/selectors.py
isidentical: Thanks @chamini2 for the information. During standup, we have decided to just copy this over since it is a relatively simple code and is unlikely to change in the short term.

src/fal/cli/selectors.py
```diff
@@ -58,26 +59,57 @@ def create_plan_from_graph(cls, parsed, nodeGraph: NodeGraph, fal_dbt: FalDbt):
...
+    components: List[str]
+
+
+def parse_union(
+    components: List[str],
+) -> SelectionUnion:
+    # Based on the original implemention at dbt-core.
+
+    # turn ['a b', 'c'] -> ['a', 'b', 'c']
+    raw_specs = itertools.chain.from_iterable(r.split(OP_SET_UNION) for r in components)
```
mederka, line 78: 😮

src/fal/cli/selectors.py
```diff
@@ -58,26 +59,57 @@ def create_plan_from_graph(cls, parsed, nodeGraph: NodeGraph, fal_dbt: FalDbt):
...
-        output.extend(set.intersection(*selector_outputs))
+    union = parse_union(selectors)
+    for intersection in union.components:
+        plan_outputs = [
+            set(SelectorPlan(selector, unique_ids, fal_dbt).execute(nodeGraph))
+            for selector in intersection.components
+            if selector
+        ]
+        if plan_outputs:
+            output |= set.intersection(*plan_outputs)
```
mederka, line 110: What's happening here?

mederka (approved):
This is beautiful! I need to catch up on the `itertools` library, it seems very powerful.

src/fal/cli/selectors.py
isidentical: We are taking the intersection of outputs, and then union'ing it into the existing result set. For example in the case of `a,b c,d`, we'll first take the intersection of `a` and `b` and then union this with the existing result set (empty set). And then in the second iteration we'll take the intersection of `c` and `d` and union this with the existing result set (intersection of `a` and `b`) which would give us `a,b` + `c,d`.

mederka: Thanks for clarification.

PR #365, opened 2022-06-08
tests: fix GitHub mutex check

chamini2:
It was running on ALL, since when profile was duckdb, it was different from postgres.


turbo1912 (approved):

chamini2:
https://github.com/fal-ai/fal/runs/6798417768?check_suite_focus=true

PR #367, opened 2022-06-08
tests: Notify also cancelled tests and change check to schedule

chamini2:
- [x] once #366 is merged, rebase and run integration tests


chamini2:
tested and it did notify: https://github.com/fal-ai/fal/runs/6803237221?check_suite_focus=true
and where it skips because it is not _scheduled_: https://github.com/fal-ai/fal/runs/6803224485?check_suite_focus=true

PR #369, opened 2022-06-08
fix: Replace table instead of dropping and renaming for BigQuery

chamini2:
This avoids getting a 404 error and then creating, which sometimes is cached by the BigQuery service and then reports a 404 even when the table has just been created.

https://github.com/googleapis/google-cloud-go/issues/975#issuecomment-383294334
> I think you answered your own question. Creating a table is an eventually consistent operation: it may not appear to exist everywhere until after some time has passed. In the screenshots above, the second request, which happens 2 seconds after the first, may end up at a different location where the creation hasn't yet propagated.

### Disclaimer
I think this works. There is no real way to test but to release it to users and await complaints.

<details>
closes FEA-156
</details>

### DIFF
--- a/src/faldbt/lib.py
+++ b/src/faldbt/lib.py
@@ -1,5 +1,4 @@
 # NOTE: INSPIRED IN https://github.com/dbt-labs/dbt-core/blob/43edc887f97e359b02b6317a9f91898d3d66652b/core/dbt/lib.py
-from time import sleep
 import six
 import os
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


chamini2:
https://github.com/fal-ai/fal/actions/runs/2464595265

.github/workflows/test_profiles.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        profile: ["postgres", "bigquery", "snowflake", "redshift", "duckdb"]
```
mederka, line 15: Might need to bring this back before merging.

mederka (changes requested):
Great solution. Much better than setting a timer.

chamini2:
https://github.com/fal-ai/fal/actions/runs/2469745557

mederka (approved):
Looks good!

PR #373, opened 2022-06-09
tests: allow testing the execution/calculation order of nodes

isidentical:
- Added support for the `Then the following nodes are calculated in the following order` condition
  - This internally collects all DBT models and their calculation (execution) time, as well as the the modification time of the Fal artifacts (from Python models or Python scripts). Which is then leveraged into sorting the list of calculated nodes.
 - Also does some refactorings on the collection of models/scripts, unifies most of the logic into a single core function we need to adapt if we change the format and a couple upper level functions.

### DIFF
--- a/integration_tests/features/flow_run_order.feature
+++ b/integration_tests/features/flow_run_order.feature
@@ -1,0 +1,42 @@
+Feature: `flow run` command with ensured order
+  Background: Project Setup
+    Given the project 002_jaffle_shop
+    When the data is seeded
+
+  Scenario: fal flow run order without --experimental-flow
+    When the following command is invoked:
+      """
+      fal flow run --profiles-dir $profilesDir --project-dir $baseDir --select middle_1+
+      """
+    Then the following nodes are calculated in the following order:
+      | middle_1 | middle_2 | after_middle | middle_2.middle_script.py |
+
+  Scenario: fal flow run order with --experimental-flow
+    When the following command is invoked:
+      """
+      fal flow run --profiles-dir $profilesDir --project-dir $baseDir --experimental-flow --select middle_1+
+      """
+    Then the following nodes are calculated in the following order:
+      | middle_1 | middle_2 | middle_2.middle_script.py | after_middle |
+
+  Scenario: fal flow run order with pure python models
+    Given the project 008_pure_python_models
+    When the data is seeded
+
+    When the following command is invoked:
+      """
+      fal flow run --profiles-dir $profilesDir --project-dir $baseDir --experimental-models
+      """
+    Then the following nodes are calculated in the following order:
+      | model_a | model_a.after.py | model_b | model_c.py | model_c.after.py | model_d | model_e.ipynb | model_e.after.py |
+
+  Scenario: fal flow run order with pure python models on a selection
+    Given the project 008_pure_python_models
+    When the data is seeded
+
+    When the following command is invoked:
+      """
+      fal flow run --profiles-dir $profilesDir --project-dir $baseDir --experimental-models --select model_c
+      """
+    Then the following nodes are calculated in the following order:
+      | model_c.py |
--- a/integration_tests/features/steps/fal_steps.py
+++ b/integration_tests/features/steps/fal_steps.py
@@ -1,14 +1,25 @@
 from functools import reduce
 import os
-from typing import Dict
+from typing import Dict, List
 from behave import *
 from fal.cli import cli
 import tempfile
 import json
 import unittest
 from os.path import exists
+from enum import Enum
 from pathlib import Path
+from datetime import datetime, timezone
 import re
+
+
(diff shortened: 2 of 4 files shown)


integration_tests/features/steps/fal_steps.py
```diff
@@ -203,35 +206,62 @@ def _temp_dir_path(context, file):
...
+
+def _get_fal_scripts(context):
+    return _unpack_dated_result(_get_dated_fal_artifacts(context, FalArtifact.SCRIPT))
+
+
+def _unpack_dated_result(dated_result):
+    if not dated_result:
+        return []
+
+    result, _ = zip(*sorted(dated_result, key=lambda node: node[1]))
```
mederka, line 228: What is zip doing here?

integration_tests/features/steps/fal_steps.py
```diff
@@ -182,19 +183,21 @@ def no_scripts_are_run(context):
...
 
 
+@then("the following nodes are calculated in the following order")
+def check_model_results(context):
+    models = _get_dated_dbt_models(context)
+    scripts = _get_dated_fal_artifacts(context, FalArtifact.MODEL, FalArtifact.SCRIPT)
+
+    ordered_nodes = _unpack_dated_result(models + scripts)
+    expected_nodes = list(map(_script_filename, context.table.headings))
+    unittest.TestCase().assertEqual(ordered_nodes, expected_nodes)
```
mederka, line 198: Nice! 👌

mederka (approved):
This is awesome. I have only minor comments.

integration_tests/features/steps/fal_steps.py
```diff
@@ -7,10 +7,22 @@
...
 from pathlib import Path
+from datetime import datetime, timezone
 import re
 
 
+class FalArtifact(int, Enum):
+    # The main distinction we can use on an artifact file to determine
+    # whether it was created by a Python script or a Python model is the number
+    # of suffixes it has. Models use <model_name>.txt and scripts use
+    # <model_name>.<script_name>.txt
```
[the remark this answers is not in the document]
isidentical: I think so. This restriction initially introduced in #360, in this PR I've refactored distinct artifact collections into a single function to unify the behavior.

mederka (approved):

integration_tests/features/flow_run_order.feature
```diff
@@ -0,0 +1,42 @@
...
+    Then the following nodes are calculated in the following order:
+      | middle_1 | middle_2 | after_middle | middle_2.middle_script.py |
+
+  Scenario: fal flow run order with --experimental-flow
+    When the following command is invoked:
+      """
+      fal flow run --profiles-dir $profilesDir --project-dir $baseDir --select middle_1+
+      """
+    Then the following nodes are calculated in the following order:
+      | middle_1 | middle_2 | after_middle | middle_2.middle_script.py |
```
chamini2, line 20:
```suggestion
      | middle_1 | middle_2 | middle_2.middle_script.py | after_middle |
```

right?

integration_tests/features/steps/fal_steps.py
```diff
@@ -203,35 +206,62 @@ def _temp_dir_path(context, file):
...
+
+def _get_fal_scripts(context) -> List[str]:
+    return _unpack_dated_result(_get_dated_fal_artifacts(context, FalArtifact.SCRIPT))
+
+
+def _unpack_dated_result(dated_result) -> List[str]:
+    if not dated_result:
+        return []
+
+    result, _ = zip(*sorted(dated_result, key=lambda node: node[1]))
```
chamini2, line 228: This is crazy, though. Good one!

integration_tests/features/steps/fal_steps.py
```diff
@@ -1,16 +1,28 @@
...
 
 
+class FalArtifact(int, Enum):
+    # The main distinction we can use on an artifact file to determine
+    # whether it was created by a Python script or a Python model is the number
+    # of suffixes it has. Models use <model_name>.txt and scripts use
+    # <model_name>.<script_name>.txt
+
+    MODEL = 1
+    SCRIPT = 2
```
chamini2, line 23: I know we are just doing tests... But this feels hacky for me a little bit. Enum values are not normally used to check against like we are doing for the artifacts' `len(artifact.suffices) in kinds`?

chamini2 (changes requested):

integration_tests/features/steps/fal_steps.py
isidentical: It actually doesn't have to be an int enum, just thought it would be better to encapsulate in a different namespace (`FalArtifact`) though no strong opinions. WDYT about the current version (just two top-level constants, `FAL_MODELS` and `FAL_SCRIPTS`)?

chamini2 (approved):
Cool!

PR #382, opened 2022-06-10
fix: docker hub deployment

mederka:


chamini2 (commented):
should we be deploying on `release` branch push?

mederka:
I think we can keep it separate and manual for now

PR #385, opened 2022-06-13
feat: post hook in fal run

mederka:
This is only for `fal run`. For `fal flow run` see [here](#386)

With this yaml:

```
version: 2

models:
  - name: model_a
    meta:
      fal:
        post-hook:
          - hook.py
          - hook2.py
```

```
dbt run
fal run
```

Will run hook.py and hook2.py. Both `write_to_source` and `write_to_model` are not allowed in these.

Closes FEA-177
Closes FEA-167

Question for @chamini2, @burkaygur and @turbo1912:

Should we allow script selection for post-hooks?

PR #386, opened 2022-06-13
feat: enable post hooks in fal flow run

mederka:
Depends on #385


src/fal/node_graph.py
```diff
@@ -58,6 +58,32 @@ def _add_after_scripts(
...
+        FalScript(faldbt, model, path, True) for path in model.get_post_hook_paths()
+    )
+    post_hook_nodes = (
+        ScriptNode(
+            _script_id_from_path(post_hook.path, model.name, "POST_HOOK"), post_hook
+        )
+        for post_hook in post_hooks
+    )
+    for post_hook_node in post_hook_nodes:
+        graph.add_node(post_hook_node.unique_id)
```
isidentical, line 79: Would it makes sense to reduce the number of for-loops here and unroll them into a single loop since each path maps to one node in the graph?

```suggestion
    for path in model.get_post_hook_paths():
        post_hook = FalScript(faldbt, model, path, True)
        script_id = _script_id_from_path(post_hook.path, model.name, "POST_HOOK")
        post_hook_node = ScriptNode(script_id, post_hook)
        graph.add_node(post_hook_node.unique_id)
 ```

File: src/fal/node_graph.py [hunk 2]
```diff
@@ -58,6 +58,32 @@ def _add_after_scripts(
...
+        nodeLookup[post_hook_node.unique_id] = post_hook_node
+        # model_fal_node depends on fal_script_node
```
isidentical, line 81: Isn't this the exact opposite in reality though? The post-hook depends on the fal node if I understood the implementation correctly.

src/fal/node_graph.py
mederka: Good catch!

src/fal/node_graph.py
mederka: Yeah, you're right

src/fal/cli/flow_runner.py
```diff
@@ -89,15 +89,27 @@ def _run_sub_graph(
...
+                    FalScript(fal_dbt, mnode.model, path, True)
+                    for path in mnode.model.get_post_hook_paths()
+                ]
+                post_hooks.extend(model_hooks)
         if len(fal_nodes) != 0:
             python_node_scripts = _id_to_fal_scripts(node_graph, fal_dbt, fal_nodes)
             results = run_scripts(python_node_scripts, fal_dbt)
             raise_for_run_results_failures(python_node_scripts, results)
 
+        if len(post_hooks):
```
turbo1912, line 109:
```
if post_hooks:
```
or 
```
if len(post_hooks) != 0 
```
right? Sorry I wrote the `if len(post_hooks) !=0 ` all over this file that is not super idiomatic python

turbo1912 (commented):
Don't you need to also change the `critical_node` calculation to take into account post_hooks?

src/fal/cli/flow_runner.py
```diff
@@ -89,15 +89,27 @@ def _run_sub_graph(
         _mark_dbt_nodes_status(fal_dbt, NodeStatus.Success, dbt_nodes)
```
chamini2, line 89: This needs to change to mark with real results. Also, we need to not raise errors until AFTER the post hooks.

src/fal/node_graph.py
```diff
@@ -178,12 +178,19 @@ def is_python_model(id: str):
                 return inode.model.python_model
             return False
 
+        def has_post_hooks(node_id: str):
+            inode = self.get_node(node_id)
+            if isinstance(inode, DbtModelNode):
+                return inode.model.get_post_hook_paths()
```
chamini2, line 184:
```suggestion
                return bool(inode.model.get_post_hook_paths())
```

chamini2 (approved):
Add a TODO regarding raising errors.

PR #391, opened 2022-06-14
fix: use internal compile_sql to avoid generating repeated names

chamini2:
image

This is the main problem:

https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/lib.py#L96

And I think it could also fix this other error which may happen because of the clash of names:

image


src/faldbt/lib.py
```diff
@@ -144,6 +144,37 @@ def _get_target_relation(
...
+        )
+
+    manifest = parse.get_dbt_manifest(config)
+    block_parser = SqlBlockParser(
+        project=config,
+        manifest=manifest,
+        root_project=config,
+    )
+
+    name = "compile_sql:" + str(hash(str(sql))) + ":" + str(uuid4())
```
mederka, line 171: Makes sense!

mederka (approved):

isidentical (approved):

PR #402, opened 2022-06-16
feat: Add `--experimental-threads=` flag

isidentical:
Only adds the flag (doesn't add anything in terms of the implementation) and splits the `fal flow run` into two isolated parts (so we can plug the threaded executor once we need it).

### DIFF
--- a/src/fal/cli/args.py
+++ b/src/fal/cli/args.py
@@ -136,6 +136,17 @@ def _add_experimental_python_models_option(parser: argparse.ArgumentParser):
     )
 
 
+def _add_experimental_threading(parser: argparse.ArgumentParser):
+    parser.add_argument(
+        "--experimental-threads",
+        type=int,
+        # TODO: Until we have all the components that can make this work
+        # we'll hide this option from --help output (only for us
+        # to know).
+        help=argparse.SUPPRESS,
+    )
+
+
 def _build_dbt_selectors(sub: argparse.ArgumentParser):
 
     # fmt: off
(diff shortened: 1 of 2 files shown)


src/fal/cli/args.py
```diff
@@ -136,6 +136,17 @@ def _add_experimental_python_models_option(parser: argparse.ArgumentParser):
     )
 
 
+def _add_experimental_threading(parser: argparse.ArgumentParser):
+    parser.add_argument(
+        "--experimental-threads",
```
isidentical, line 141: Note: We might call this `--threads=...` in the future (or `--workers` etc.), this is just the name for us (to test).

turbo1912 (approved):

src/fal/cli/args.py
chamini2: We already have a `--threads` flag

```
  --threads THREADS     Specify number of threads to use while executing Python scripts and dbt models. Overrides settings in profiles.yml.
```

isidentical: Ah, good to know! We might use this and maybe change the description. It depends whether we want to enable the new parallelization code (and the complete refactor of the stuff) by default on the initial release though.

src/fal/cli/flow_runner.py
```diff
@@ -19,31 +19,59 @@
...
+    if _is_experimental_models_enabled(parsed):
+        telemetry.log_call("experimental_python_models")
+        generated_models = generate_python_dbt_models(parsed.project_dir)
+
+    fal_dbt = create_fal_dbt(parsed, generated_models)
+    _mark_dbt_nodes_status(fal_dbt, NodeStatus.Skipped)
+
+    node_graph = NodeGraph.from_fal_dbt(fal_dbt)
+    if parsed.experimental_threads is not None:
+        run_threaded(fal_dbt=fal_dbt, parsed=parsed, node_graph=node_graph)
```
chamini2, line 71: Can you add `telemetry.log_call("experimental_threads")` before running it.

chamini2 (approved):

mederka (approved):

PR #403, opened 2022-06-16
feat: Implement the basic planner

isidentical:
A port of the Gorkem's algorithm (with a few modifications, like ancestoral checks) to work on raw `nx.DiGraph` objects and reduce the subgraphs in place.
image


src/fal/planner/plan.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def split() -> Iterator[list[str]]:
+        if len(current_stack) > 1:
+            yield current_stack.copy()
+        current_stack.clear()
+        allowed_ancestors.clear()
+
+    for node in nx.topological_sort(graph):
+        properties = graph.nodes[node]
+        if properties["kind"] == "python model":
```
isidentical, line 30: We currently don't take scripts into account.

src/fal/planner/plan.py
```diff
@@ -0,0 +1,77 @@
...
```
turbo1912, line 29: small question about this one; why are you assuming this to exist? dont we need a more structured node representation first? similar to what's discussed in the doc 

```
@dataclass
FalPythonModel(FalWorkUnit)
  dbt_model: ???
	script: FalScript
	hooks: List[FalHook]

@dataclass
DbtModel(FalWorkUnit)
	models: List[str]
	hooks: List[FalHook]

@dataclass
FalHook(FalWorkUnit)
	script: FalScript
```

isidentical: > small question about this one; why are you assuming this to exist? dont we need a more structured node representation first? similar to what's discussed in the doc

We convert to this representation when changing it from raw graph nodes into tasks at the scheduler. I think it is a bit easier to traverse/manipulate with the existing form on the planner. I think a better option would be ensuring these are issued at the graph construction stage, I'll also bundle the changes needed to the node graph to this PR. That should make this work, WDYT?

src/fal/node_graph.py
```diff
@@ -102,11 +109,15 @@ def from_fal_dbt(cls, fal_dbt: FalDbt):
         for model in fal_dbt.list_models():
             model_fal_node = DbtModelNode(model.unique_id, model)
             node_lookup[model_fal_node.unique_id] = model_fal_node
-            graph.add_node(model_fal_node.unique_id)
+            graph.add_node(
+                model_fal_node.unique_id,
+                kind=NodeKind.FAL_MODEL if model.python_model else NodeKind.DBT_MODEL,
+                **{"post-hook": model.get_post_hook_paths()},
```
chamini2, line 115: why not `post_hook`

src/fal/planner/plan.py
```diff
@@ -0,0 +1,77 @@
...
+
+from typing import Iterator
+
+import networkx as nx
+
+
+def _find_subgraphs(graph: nx.DiGraph) -> Iterator[list[str]]:
+    # Initially topologically sort the graph, and split it
+    # into chunks by using the critical node approach (post-hooks
+    # are also a thing now).
```
chamini2, line 11: I think the 
> (post-hooks are also a thing now)

could stay there after post-hooks are a normal thing

Maybe update with a list of things we handle

```
- post-hooks
- after-scripts
- before-scripts
- ...
```

or don't put this part of the comment

src/fal/planner/plan.py
chamini2, line 29:
```suggestion
        if properties["kind"] == NodeKind.FAL_MODEL:
```

There is no `"python model"` kind.

src/fal/planner/plan.py
```diff
@@ -0,0 +1,77 @@
...
+    # Then for each chunk, check whether they have the
+    # same set of ancestors. This is important because
+    # we want to group only the things that are in
+    # the same branch.
+
+    current_stack = []
+    allowed_ancestors = set()
+
+    def split() -> Iterator[list[str]]:
+        if len(current_stack) > 1:
```
chamini2, line 22:
```suggestion
        if len(current_stack) >= 1:
```

src/fal/planner/plan.py
```diff
@@ -0,0 +1,77 @@
...
+        if properties["kind"] == "python model":
+            yield from split()
+            continue
+
+        ancestors = nx.ancestors(graph, node)
+        if not current_stack:
+            allowed_ancestors = ancestors
+
+        if not ancestors.issubset(allowed_ancestors):
+            yield from split()
```
chamini2, line 38: No need to `continue`?

chamini2 (commented):
I need some more time to read more carefully. I do not completely understand yet.

src/fal/planner/plan.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # Then for each chunk, check whether they have the
+    # same set of ancestors. This is important because
+    # we want to group only the things that are in
+    # the same branch.
+
+    current_stack = []
+    allowed_ancestors = set()
+
+    def split() -> Iterator[list[str]]:
```
chamini2, line 21: yields 0 or 1 items, sort of like a Haskell `Maybe` in Python... Interesting usage! Also reminded me of js observables

src/fal/planner/plan.py
```diff
@@ -0,0 +1,77 @@
+from __future__ import annotations
+
+from typing import Iterator
+
+import networkx as nx
+
+
+def _find_subgraphs(graph: nx.DiGraph) -> Iterator[list[str]]:
```
chamini2, line 8:
```py
from typing import List
```

We support Python 3.7

src/fal/planner/plan.py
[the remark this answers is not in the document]
isidentical: Awesome text, thanks!

src/fal/node_graph.py
```diff
@@ -102,11 +109,15 @@ def from_fal_dbt(cls, fal_dbt: FalDbt):
         for model in fal_dbt.list_models():
             model_fal_node = DbtModelNode(model.unique_id, model)
             node_lookup[model_fal_node.unique_id] = model_fal_node
-            graph.add_node(model_fal_node.unique_id)
+            graph.add_node(
+                model_fal_node.unique_id,
+                kind=NodeKind.FAL_MODEL if model.python_model else NodeKind.DBT_MODEL,
+                post_hook=model.get_post_hook_paths(),
+            )
```
mederka, line 116: Cool

src/fal/planner/plan.py
```diff
@@ -0,0 +1,75 @@
...
+        if properties["kind"] is NodeKind.FAL_MODEL:
+            yield from split()
+            continue
+
+        ancestors = nx.ancestors(graph, node)
+        if not current_stack:
+            allowed_ancestors = ancestors
+
+        if not ancestors.issubset(allowed_ancestors):
+            yield from split()
```
mederka, line 36: Why don't we continue here?

src/fal/planner/plan.py
```diff
@@ -0,0 +1,75 @@
...
+
+        current_stack.append(node)
+        allowed_ancestors |= {node, *ancestors}
```
mederka, line 39: What's happening here?

tests/planner/test_plan.py
```diff
@@ -0,0 +1,19 @@
...
+    graph = graph_info["graph"]
+    new_graph = plan_graph(graph)
+
+    subgraphs = [
+        maybe_subgraph
+        for maybe_subgraph in new_graph.nodes
+        if isinstance(maybe_subgraph, nx.DiGraph)
+    ]
+
+    assert {frozenset(subgraph.nodes) for subgraph in subgraphs} == graph_info["subgraphs"]
```
mederka, line 19: This is great!

mederka (commented):
Great job, @isidentical!

I just have a couple of noob questions.

src/fal/planner/plan.py
isidentical: We don't need it, if the ancestor check fails we are detaching from the current stack and starting a new one.

tests/planner/test_plan.py
isidentical: Thanks!

src/fal/planner/plan.py
isidentical: We do use `continue` in the Python models because it is theoretically impossible for them to be the start of a new stack (since they would cut off immediately).

src/fal/planner/plan.py
chamini2: It is a union operation of sets

mederka (approved):
Thanks for answering the question!

PR #404, opened 2022-06-16
feat: Add source freshness awareness

### LINKED ISSUE #390: support Source Freshness
**Context**
What are you trying to do and how would you want to do it differently? Is it something you currently cannot do?

after I run `dbt source freshness`, have fal be aware of the results and (possibly) run python scripts based on the results
**Is your feature request related to a problem? Please describe.**

**Additional context**

chamini2:
closes #390


src/faldbt/parse.py
```diff
@@ -103,6 +103,25 @@ def get_dbt_manifest(config) -> Manifest:
...
+    try:
+        # BACKWARDS: Change intorduced in 1.0.0
+        if hasattr(FreshnessExecutionResultArtifact, "read_and_check_versions"):
+            return FreshnessExecutionResultArtifact.read_and_check_versions(
+                sources_path
+            )
+        else:
+            return FreshnessExecutionResultArtifact.read(sources_path)
+
+    except IncompatibleSchemaException as exc:
```
mederka, line 117: We should have a test where this is raised

src/faldbt/parse.py
```diff
@@ -103,6 +103,25 @@ def get_dbt_manifest(config) -> Manifest:
...
+        exc.add_filename(sources_path)
+        raise
+    except RuntimeException as exc:
```
mederka, line 120: In what cases does this happen?

src/faldbt/project.py
```diff
@@ -169,30 +175,46 @@ def get_models(self) -> List[DbtModel]:
...
+    @property
+    def results(self) -> Sequence[RunResultOutput]:
         if self.nativeRunResult:
-            self.results = nativeRunResult.results
+            return self.nativeRunResult.results
+        else:
+            return []
+
+
+@dataclass(init=False)
```
mederka, line 205: Not sure if this needs to be a dataclass

src/faldbt/project.py
```diff
@@ -683,33 +712,51 @@ def _firestore_dict_to_document(data: Dict, key_column: str):
 
 
 def _map_nodes(
-    run_results: DbtRunResult, manifest: DbtManifest, generated_models: Dict[str, Path]
+    run_results: DbtRunResult,
```
mederka, line 715: Is it too much to ask to refactor this function? Even splitting it into two would be helpful.

mederka (commented):
Great work, @chamini2! I have some comments. Open to discussion.

src/faldbt/project.py
isidentical: Maybe we can remove the `__init__` and remove the `init=False` from the `@dataclass`? The `__init__` doesn't seem to add any value as is.

src/faldbt/project.py
chamini2: It was merged before this PR. I separated them because now we have
- better type reporting in editor
- test check changes in the next PR for sources (https://github.com/fal-ai/fal/pull/405/files#diff-ba8ca12dab9905543efcdff4c5776ab3af382f7e6bd015cf68fb31032b8909bbR768)
- I think it may be better to refactor like Mederis suggesting, since we create nodes and then assign statuses

src/faldbt/parse.py
chamini2: when the `target/sources.json` file is not there

chamini2:
checks did not trigger: 

- https://github.com/fal-ai/fal/actions/runs/2537040560
- https://github.com/fal-ai/fal/actions/runs/2537039827

src/faldbt/project.py
```diff
@@ -80,28 +92,39 @@ def __post_init__(self):
 
 
 @dataclass
-class DbtSource(_DbtNode):
-    tests: List[DbtTest] = field(init=False, default_factory=list)
-    status: str = field(init=False, default=None)
+class DbtTestableNode(_DbtNode):
```
mederka, line 95: nit: _DbtTestableNode

src/faldbt/project.py
```diff
@@ -80,28 +92,39 @@ def __post_init__(self):
...
+    tests: List[DbtTest] = field(default_factory=list)
+
+    def _get_status(self):
+        if self._status == NodeStatus.Skipped and any(
+            map(lambda test: test != NodeStatus.Skipped, self.tests)
```
mederka, line 100:
```suggestion
            (test != NodeStatus.Skipped for test in self.tests)
```

src/faldbt/project.py
```diff
@@ -171,48 +194,98 @@ def get_scripts(self, keyword: str, before: bool) -> List[str]:
...
+        generated_models: Dict[str, Path],
+    ) -> Tuple[List[DbtModel], List[DbtSource], List[DbtTest]]:
+        status_map = dict(
             map(
-                lambda node: DbtTest(node),
-                filter(
-                    lambda node: node.resource_type == NodeType.Test,
-                    self.nativeManifest.nodes.values(),
-                ),
+                lambda res: (res.unique_id, res.status),
```
mederka, line 248: nit: I really prefer list comprehension to lambda expressions

src/faldbt/project.py
```diff
@@ -171,48 +194,98 @@ def get_scripts(self, keyword: str, before: bool) -> List[str]:
...
+        tests: List[DbtTest] = []
+        tests_dict: Dict[str, List[DbtTest]] = defaultdict(list)
+        for node in self.get_test_nodes():
+            test = DbtTest(
+                node=node,
+                _status=status_map.get(node.unique_id, NodeStatus.Skipped).value,
+            )
+            tests.append(test)
 
+            tests_dict[test.model].append(test)
```
mederka, line 262: I like this

mederka (commented):
👏 I think it's definitely going in the right direction. Good idea with mapping tests to a dict first. As mentioned in the comments, maybe it's better to use list comprehension rather that `map` function. We can even make generators, if we use () instead of [].

src/faldbt/project.py
```diff
@@ -171,48 +194,93 @@ def get_scripts(self, keyword: str, before: bool) -> List[str]:
...
+    def get_source_nodes(self) -> Iterable[ParsedSourceDefinition]:
+        return self.nativeManifest.sources.values()
 
+    def _map_nodes(
+        self,
+        run_results: DbtRunResult,
+        freshness_results: DbtFreshnessExecutionResult,
+        generated_models: Dict[str, Path],
+    ) -> Tuple[List[DbtModel], List[DbtSource], List[DbtTest]]:
+        status_map = dict((r.unique_id, r.status) for r in run_results.results)
```
isidentical, line 248: nit: maybe let's start using dict comprehensions
```suggestion
        status_map = {r.unique_id: r.status for r in run_results.results}
 ```

**Review by isidentical — APPROVED** [review]
Amazing job, everything looks great 🚀

File: src/faldbt/project.py [hunk 10]
```diff
@@ -171,48 +194,93 @@ def get_scripts(self, keyword: str, before: bool) -> List[str]:
...
         )
 
-    def get_sources(self) -> List[ParsedSourceDefinition]:
-        return list(
-            map(lambda node: DbtSource(node=node), self.nativeManifest.sources.values())
-        )
+    def get_source_nodes(self) -> Iterable[ParsedSourceDefinition]:
+        return self.nativeManifest.sources.values()
 
+    def _map_nodes(
```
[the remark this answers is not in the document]
chamini2: I think we can keep it like this for now and if we have good reasons to, we modify it later.

PR #405, opened 2022-06-16
fix: Make tests for sources to be picked up correctly

chamini2:
<details>
closes FEA-192
</details>


src/faldbt/project.py
```diff
@@ -65,25 +65,43 @@ def set_status(self, status: str):
...
+                if refs:
+                    self.model = refs[0]
+
+                print(metadata.kwargs["model"])
+                sources = re.findall(
+                    r"source\('([^']+)', *'([^']+)'\)", metadata.kwargs["model"]
+                )
+                print(sources)
+                if sources:
+                    self.source = sources[0]
```
mederka, line 96: What does this tuple look like?

src/faldbt/project.py
```diff
@@ -65,25 +65,43 @@ def set_status(self, status: str):
...
-                self.column = node.test_metadata.kwargs.get("column_name", None)
+                self.column = metadata.kwargs.get("column_name", None)
+
+                # TODO: Handle package models?
+                refs = re.findall(r"ref\('([^']+)'\)", metadata.kwargs["model"])
+
+                if refs:
+                    self.model = refs[0]
+
+                print(metadata.kwargs["model"])
```
mederka, line 90: Probably want to remove print statements

src/faldbt/project.py
```diff
@@ -65,25 +65,43 @@ def set_status(self, status: str):
...
+                sources = re.findall(
```
mederka, line 91: So subquery can have `source` as well? Sorry didn't see this.

mederka (commented):
Thanks you for looking into this, @chamini2! The PR looks good, just small questions.

src/faldbt/project.py
chamini2: `sources` looks like
```
[('results', 'some_source')]
```

mederka (approved):

src/faldbt/project.py
```diff
@@ -250,15 +266,18 @@ def _map_nodes(
...
                 node=node,
                 _status=status_map.get(node.unique_id, NodeStatus.Skipped).value,
             )
             tests.append(test)
 
-            tests_dict[test.model].append(test)
+            if test.model:
+                tests_dict[test.model].append(test)
+            if test.source:
+                tests_dict[test.source].append(test)
```
chamini2, line 280: curious on opinions here

isidentical: I think it looks good!

PR #406, opened 2022-06-17
feat: Implement basic task units for parallelization

isidentical:
Depends on #403

### DIFF
--- a/src/fal/planner/tasks.py
+++ b/src/fal/planner/tasks.py
@@ -1,0 +1,158 @@
+from __future__ import annotations
+
+import argparse
+import sys
+import traceback
+import uuid
+from contextlib import contextmanager
+from dataclasses import dataclass, field
+from functools import cached_property
+from typing import Iterator, List, Union, Any
+
+from dbt.logger import GLOBAL_LOGGER as logger
+
+from fal.node_graph import FalScript
+from fal.utils import print_run_info
+from faldbt.project import DbtModel, FalDbt, NodeStatus
+
+SUCCESS = 0
+FAILURE = 1
+
+
+class Task:
+    def execute(self, args: argparse.Namespace, fal_dbt: FalDbt) -> int:
+        raise NotImplementedError
+
+
+def _unique_id_to_model_name(unique_id: str) -> str:
+    split_list = unique_id.split(".")
+    # if its a unique id 'model.fal_test.model_with_before_scripts'
+    return split_list[len(split_list) - 1]
+
+
+def _unique_ids_to_model_names(id_list: List[str]) -> List[str]:
+    return list(map(_unique_id_to_model_name, id_list))
+
+
+def _mark_dbt_nodes_status(
+    fal_dbt: FalDbt,
+    status: NodeStatus,
+    dbt_node: Union[str, None] = None,
+):
+    for model in fal_dbt.models:
+        if dbt_node is not None:
+            if model.unique_id == dbt_node:
+                model.set_status(status)
+        else:
+            model.set_status(status)
+
+
+def _map_cli_output_model_statuses(
+    run_results: dict[Any, Any]
+) -> Iterator[tuple[str, NodeStatus]]:
+    if not isinstance(run_results.get("results"), list):
+        raise Exception("Could not read dbt run results")
+
+    for result in run_results["results"]:
+        if not result.get("unique_id") or not result.get("status"):
+            continue
+
+        yield result["unique_id"], NodeStatus(result["status"])
+
+
+def _run_script(script: FalScript, fal_dbt: FalDbt):
+    print_run_info([script])
+    logger.debug("Running script {}", script.id)
+    try:
+        with _modify_path(fal_dbt):
+            script.exec(fal_dbt)
+    except:
+        logger.error("Error in script {}:\n{}", script.id, traceback.format_exc())
+        # TODO: what else to do?
+        status = FAILURE
+    else:
+        status = SUCCESS
+    finally:
+        logger.debug("Finished script {}", script.id)
+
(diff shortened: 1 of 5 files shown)


src/fal/planner/tasks.py
```diff
@@ -0,0 +1,159 @@
...
+from typing import Iterator, Any
+
+from dbt.logger import GLOBAL_LOGGER as logger
+
+from fal.node_graph import FalScript
+from fal.utils import print_run_info
+from faldbt.project import DbtModel, FalDbt, NodeStatus
+
+SUCCESS = 0
+FAILURE = 1
```
chamini2, line 19: should this be an enum?

src/fal/planner/tasks.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def _unique_ids_to_model_names(id_list: list[str]) -> list[str]:
+    return list(map(_unique_id_to_model_name, id_list))
+
+
+def _mark_dbt_nodes_status(
+    fal_dbt: FalDbt,
+    status: NodeStatus,
+    dbt_node: str | None = None,
```
chamini2, line 40: We support Python 3.7

src/fal/planner/tasks.py
```diff
@@ -0,0 +1,159 @@
...
+        logger.debug("Finished script {}", script.id)
+
+    return status
+
+
+@contextmanager
+def _modify_path(fal_dbt: FalDbt):
+    sys.path.append(fal_dbt.scripts_dir)
+    yield
+    sys.path.remove(fal_dbt.scripts_dir)
```
chamini2, line 85: Beautiful. the `remove` happens even if there was an error in the `yield`?

src/fal/planner/tasks.py
```diff
@@ -0,0 +1,159 @@
...
+
+        status = NodeStatus.Success if script_result == SUCCESS else NodeStatus.Error
+        _mark_dbt_nodes_status(fal_dbt, status, self.bound_model.unique_id)
+        return script_result
+
+
+@dataclass
+class FalHookTask(Task):
+    hook_path: str
+    bound_model: DbtModel | None = None
```
chamini2, line 136: this cannot be really `None`, right?

src/fal/planner/tasks.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def execute(self, args: argparse.Namespace, fal_dbt: FalDbt) -> int:
+        script = FalScript(fal_dbt, self.bound_model, self.hook_path, True)
+        return _run_script(script, fal_dbt=fal_dbt)
+
+
+@dataclass
+class Node:
+    task: Task
```
chamini2, line 145: this is a `DBTTask` because `FalModelTask` extends that class. I am OK with it being `Task`, though.

src/fal/planner/tasks.py
```diff
@@ -0,0 +1,159 @@
...
+    post_hooks: list[Task] = field(default_factory=list)
```
chamini2, line 146: same here, it is really a `FalHookTask`

src/fal/planner/tasks.py
```diff
@@ -0,0 +1,159 @@
...
+    dependencies: list[Node] = field(default_factory=list)
+
+    def exit(self, status: int) -> None:
```
chamini2, line 149: status should be enum instead?

tests/planner/test_tasks.py
```diff
@@ -0,0 +1,90 @@
...
+        "fal.fal_script.FalScript.model_script",
+        return_value=FakeScript(),
+    )
+    mocker.patch(
+        "fal.planner.tasks._run_script",
+        return_value=return_code,
+    )
+
+
+@pytest.mark.parametrize("return_code", [SUCCESS, FAILURE])
```
chamini2, line 53: This is very cool

src/fal/planner/tasks.py
isidentical: I don't think it works for now (it would be basically `SUCCESS`, since anything beside it is technically an error). Maybe in the future when we implement the `SKIPPED` though.

src/fal/planner/tasks.py
isidentical: True, true (it would probably work since it is under `from __future__ import annotations` but yea let's do these when we drop 3.7).

src/fal/planner/tasks.py
isidentical: Yep, but since it is assigned afterwards just made it so to satisfy type checkers.

src/fal/planner/tasks.py
isidentical: For now, I'd abstain from that (just to see where we'd end up with these statuses, right now we only use `SUCCESS` theoretically but let's reconsider this on the integration phase).

turbo1912 (approved):
I would like us to merge some of this work in.. from what I understand and reading the PR feedback we have some questions around the plan step. the plan step being the first commit of this chain of commits it is blocking us to move forward. For the sake of making progress while we figure out what to do with plan can you change the planner to just do something really basic (maybe even just topologically sort).

src/fal/planner/tasks.py
```diff
@@ -0,0 +1,155 @@
...
+        return _run_script(script, fal_dbt=fal_dbt)
+
+
+@dataclass
+class Node:
+    task: DBTTask
+    post_hooks: List[FalHookTask] = field(default_factory=list)
+    dependencies: List[Node] = field(default_factory=list)
+
+    def exit(self, status: int) -> None:
```
chamini2, line 146: ATL: when will we call `exit`?

PR #407, opened 2022-06-17
feat: Implement the scheduler

isidentical:
Depends on #406


src/fal/planner/schedule.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def create_node(
+    node: str | nx.DiGraph, properties: dict, node_graph: NodeGraph
+) -> Node:
+    kind = properties["kind"]
+    if isinstance(node, nx.DiGraph):
+        model_ids = sorted(
+            list(node),
+            key=lambda node: node == properties["exit_node"],
```
chamini2, line 19: add a comment about sorting by boolean

src/fal/planner/schedule.py
```diff
@@ -0,0 +1,133 @@
...
+    else:
+        assert kind is NodeKind.FAL_MODEL
+        task = FalModelTask(model_ids=model_ids)
+
+    model_node = node_graph.get_node(model_ids[-1])
+    assert model_node is not None
+
+    if isinstance(task, FalModelTask):
+        assert isinstance(model_node, DbtModelNode)
+        task.bound_model = model_node.model
```
chamini2, line 35: couldn't this happen in the constructor?

src/fal/planner/schedule.py
```diff
@@ -0,0 +1,133 @@
...
+    def _stage_node(self, target_node: Node) -> None:
+        # When we start running a node, we don't want to simply remove it
+        # just yet (since that would unblock all of its dependencies). So
+        # we temporarily add it to a separate group of nodes (staged).
+
+        self._counter += 1
+        target_node.set_run_index(self._counter)
+
+        self.nodes.remove(target_node)
+        self._staged_nodes.append(target_node)
```
chamini2, line 88: Is it possible that a race condition could tink a dependency is unblocked? Should we swap the order of these 2?

src/fal/planner/schedule.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def finish(self, target_node: Node, status: int) -> None:
+        # When a staged node's execution is finished, we'll remove it
+        # alltogether and unblock all of its dependencies.
+        self._staged_nodes.remove(target_node)
+
+        if status == SUCCESS:
+            self._succeed(target_node)
+        else:
+            self._fail(target_node)
```
chamini2, line 98: should we do this before `self._staged_nodes.remove(target_node)` for race conditions?

src/fal/planner/schedule.py
```diff
@@ -0,0 +1,133 @@
...
+        for node in self.nodes.copy():
+            if target_node in node.dependencies:
+                self.nodes.remove(node)
+
+    def _succeed(self, target_node: Node) -> None:
+        for node in self.nodes.copy():
+            if target_node in node.dependencies:
+                node.dependencies.remove(target_node)
+
+    def iter_available_nodes(self) -> Iterator[Node]:
```
chamini2, line 112: Should this return a node instead of being an iterator of nodes? If it has a lot of unblocked nodes at the beginning without any depth, it will still process all of those.

```mermaid
graph LR;
  a;
  b --> c --> d --> e --> f --> g;
```

Would yield `b`, then `a`, and then need to be called again and get `c`, etc.

src/fal/planner/schedule.py
```diff
@@ -0,0 +1,133 @@
...
+            self._succeed(target_node)
+        else:
+            self._fail(target_node)
+
+        target_node.exit(status)
+
+    def _fail(self, target_node: Node) -> None:
+        for node in self.nodes.copy():
+            if target_node in node.dependencies:
+                self.nodes.remove(node)
```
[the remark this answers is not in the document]
chamini2: I see dependencies have all ancestors, not just direct dependencies.

src/fal/planner/schedule.py
```diff
@@ -0,0 +1,133 @@
...
+        # node.
+        unblocked_nodes = [node for node in self.nodes if not node.dependencies]
+        unblocked_nodes.sort(key=self._calculate_node_score, reverse=True)
+
+        for node in unblocked_nodes:
+            self._stage_node(node)
+            yield node
+
+
+def schedule_graph(graph: nx.DiGraph, node_graph: NodeGraph) -> NodeQueue:
```
chamini2, line 124: `node_graph` has a `nx.DiGraph` in `node_graph.graph`. Is the one received as first arg the same?

src/fal/planner/schedule.py
isidentical: I initially did it that way, but when I wrote the executor I've noticed that it can take as many as available so instead of calculating unblocked nodes at each call we only do calculate once and give up everything we can at the same time. If we did handle the number of jobs in the queue for the thread pool executor manually, I think your suggestion might be a decent optimization but currently it is all managed by the `concurrent.futures` so even if make this a method we'll still take all the unblocked nodes at each iteration and pass them at the same time.

isidentical: Maybe this could be a feature in the future though, with the improvements in the node scoring system (like timing info processing) it might matter a lot more. But for now, I'd abstain from that. WDYT?

src/fal/planner/schedule.py
isidentical: I am unsure about how a race can happen here. Would you mind giving an example?

src/fal/planner/schedule.py
isidentical: Same as question as below.

src/fal/planner/schedule.py
isidentical: Ah seems like I can simply lift the `model_node` retrieval to a bit above, seems to be solved now!

src/fal/planner/schedule.py
chamini2: I see now that `_staged_nodes` is not really used to check if a node is in process or not. `_staged_nodes` really does not do much, we could not have it at all. Unless it is useful for debugging, maybe.

isidentical: Yep, that is the primary use case (a bit generalized, like not making a node disappear all together until it is finished).

src/fal/planner/schedule.py
isidentical: > I say change this not to a list of nodes but to return the most scored node and that is it.

But it really doesn't solve the problem with this system. If we change this to something that only returns a single node, then we'll port this logic back to the executor which will call this method as much as it can since we don't know how many available threads are there:

```py
while (node := scheduler.get_node()) is not None:
     submit_future_group(node)
```

which would be pretty much the same with the current approach, even a bit slower and a bit more complicated. Though maybe I am missing something.

src/fal/planner/tasks.py
```diff
@@ -0,0 +1,155 @@
...
+    hook_path: str
+    bound_model: Union[DbtModel, None] = None
+
+    def execute(self, args: argparse.Namespace, fal_dbt: FalDbt) -> int:
+        script = FalScript(fal_dbt, self.bound_model, self.hook_path, True)
+        return _run_script(script, fal_dbt=fal_dbt)
+
+
+@dataclass
+class Node:
```
turbo1912, line 141: Is `Node` the combination of a `DbtTask` and its post hooks? Seems like we need to name this something more specific

src/fal/planner/schedule.py
chamini2: > then we'll port this logic back to the executor which will call this method as much as it can since we don't know how many available threads are there:
> ```
> while (node := scheduler.get_node()) is not None:
>      submit_future_group(node)
> ```
> which would be pretty much the same with the current approach, even a bit slower and a bit more complicated. Though maybe I am missing something.

Got it. I see your point! I think maybe a future change would be to make the executor aware of available threads, so it just calls when it makes sense. For now, let's leave it like it is!

Thanks for the explanation!

src/fal/planner/schedule.py
```diff
@@ -0,0 +1,132 @@
...
+        self._counter += 1
+        target_node.set_run_index(self._counter)
+
+        self.nodes.remove(target_node)
+        self._staged_nodes.append(target_node)
+
+    def finish(self, target_node: Node, status: int) -> None:
+        # When a staged node's execution is finished, we'll remove it
+        # alltogether and unblock all of its dependencies.
+        self._staged_nodes.remove(target_node)
```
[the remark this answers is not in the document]
chamini2: `_staged_nodes` really does not block stuff from starting. It is a way to keep the running nodes somewhere, but we could do without it.

isidentical:
@turbo1912 @chamini2 short update:

- Renamed `Node` to `TaskGroup` based on your feedback. I think that kind of covers its purpose (like multiple tasks grouped together).
- Renamed `NodeQueue` to `Scheduler` (and removed references to `node` name under it, so renamed a couple other methods).
- Added more explanation behind `exit_node` into the scheduler.
- Removed `_staged_nodes`, it didn't provide enough value to stay. So based on your suggestions I also removed it.

src/fal/planner/schedule.py
```diff
@@ -0,0 +1,133 @@
...
+    def finish(self, target_group: TaskGroup, status: int) -> None:
+        # When a staged group's execution is finished, we'll remove it
+        # altogether and unblock all of its dependencies.
+
+        if status == SUCCESS:
+            self._succeed(target_group)
+        else:
+            self._fail(target_group)
+
+        target_group.exit(status)
```
chamini2, line 98: Right now this stops the whole execution, right?

chamini2 (approved):

src/fal/planner/schedule.py
isidentical: Yep, we have [FEA-207](https://linear.app/features-and-labels/issue/FEA-207/implement-error-recovery-when-a-node-fails-or-independent-nodes-still) as a sub-issue specifically for the feature we talked.

PR #408, opened 2022-06-20
feat: Implement the executor 

isidentical:


src/fal/planner/executor.py
```diff
@@ -0,0 +1,101 @@
...
+            completed_futures, _ = wait(futures, return_when=FIRST_COMPLETED)
+            for future in completed_futures:
+                group = future.group
+                group.process(future)
+                if group.is_done:
+                    queue.finish(group.node, status=group.status)
+
+            # And load all the tasks that were blocked by those futures.
+            future_groups.extend(create_futures(executor))
+            futures = get_futures(future_groups)
```
chamini2, line 101: If we have more than 1 thread, and we have 2 separate `DBTTask` that are NOT dependent on each other.

Could this mean we are running two dbt tasks at the same time?

chamini2: I did an experiment in this branch and I see that we may be running 2 separate `DBTTask` at the same time. This breaks the dbt thread config dbt will try to use all the threads configured in BOTH runs.

I think a solution is to NOT group nodes at all in the `plan` stage and have a single dbt model per `DBTTask`. Since we handle parallelization ourselves, this will enable 

1. faster running of post-hooks after the corresponding dbt model
2. not break the user expectations of the number of threads being used in the data warehouse

chamini2: The thing about parallelizing dbt runs (happens in both solutions) is that https://docs.getdbt.com/reference/project-configs/target-path gets overwritten for each, meaning we have race conditions for reading the `run_results.json`

isidentical: > Could this mean we are running two dbt tasks at the same time?

Yes!

> This breaks the dbt thread config dbt will try to use all the threads configured in BOTH runs.

That is true and a bad side-effect. 

> I think a solution is to NOT group nodes at all in the plan stage and have a single dbt model per DBTTask. Since we handle parallelization ourselves, this will enable

I haven't thinked it like that, and yes it makes sense. (CC: @turbo1912 @mederka). The grouping makes a decent improvement on short running tasks (since each DBT call have a ~0.5-1 second delay) but I don't see any other way around this problem.

> The thing about parallelizing dbt runs (happens in both solutions) is that https://docs.getdbt.com/reference/project-configs/target-path gets overwritten for each, meaning we have race conditions for reading the run_results.json

I think this is another point which won't go away even if we remove the planner. I already have a hack for this in the main PR that makes it work on tests but we'll need a more permanent solution (see the comment there):
https://github.com/fal-ai/fal/pull/383/files#diff-204bc8d05abda9e5becbbb1d2e7bd91accec607176df85f2286e2d993768412dR112-R118

chamini2: Agreed with everything you said, @isidentical . Thanks for reading! 

I had not seen the hack in the tests, I see. I have asked in the dbt Slack community and someone prompted me to open a GitHub issue asking for a flag to control this. Let's see what the dbt team thinks about this.

turbo1912 (approved):

src/fal/planner/executor.py
isidentical: As we have discussed internally, let's defer the threading issue until the full integration. For the second problem, I'll send a PR that runs DBT through Python so we can get around it.

src/fal/planner/executor.py
```diff
@@ -0,0 +1,101 @@
+from __future__ import annotations
```
chamini2, line 1: ATL: what does this do?

chamini2 (approved):

src/fal/planner/executor.py
isidentical: It postpones the execution of annotations. Pretty much responsible for bringing more flexibility to annotations when needed (when I was using `list[x]` syntax or `x | y`, it was allowing these to run in 3.7 even though they are not implemented at that version). Right now, it doesn't add any value, removing it!

PR #409, opened 2022-06-20
feat: Allow Python models to be the exit nodes on subgraphs.

isidentical:
We now allow ephemeral models to act like exit nodes. Previously we only supported DBT nodes to be exit nodes, but with this patch pure Python models are also supported.


isidentical:
We are currently putting a hold on the planner stuff, closing it for now.

PR #410, opened 2022-06-20
tests: allow representing parallel executions in order tests

isidentical:
We are switching from tables to our own format, since tables in cucumber needs to have the exact same number of columns in each row which is something that doesn't quite work for us (see the examples below). This PR adds a new way to write order based tests, with the ability of specifiying whether a row might be executed in parallel.

```
    Then the following nodes are calculated in the following order:
      """
      1. model_a, model_a.after.py, model_b, model_c.py
      2. (unordered) model_c.post_hook.py, model_c.after.py
      3. (unordered) model_d, model_e.ipynb, model_e.after.py
      """
```

### DIFF
--- a/integration_tests/features/flow_run_order.feature
+++ b/integration_tests/features/flow_run_order.feature
@@ -9,7 +9,12 @@ Feature: `flow run` command with ensured order
       fal flow run --profiles-dir $profilesDir --project-dir $baseDir --select middle_1+
       """
     Then the following nodes are calculated in the following order:
-      | middle_1 | middle_2 | after_middle | middle_2.middle_script.py |
+      """
+      1. middle_1
+      2. middle_2
+      3. after_middle,
+      4. middle_2.middle_script.py
+      """
 
   Scenario: fal flow run order with --experimental-flow
     When the following command is invoked:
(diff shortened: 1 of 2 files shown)


chamini2:
>     Then the following nodes are calculated in the following order:
>       """
>       1. model_a, model_a.after.py, model_b, model_c.py
>       2. (unordered) model_c.post_hook.py, model_c.after.py
>       3. (unordered) model_d, model_e.ipynb, model_e.after.py
>       """
> ```

My only feedback is that I think is kind of weird there are 2 ways of transmitting order. The above could be also represented like:

```
"""
1. model_a
2. model_a.after.py
3. model_b
4. model_c.py
5. (unordered) model_c.post_hook.py, model_c.after.py
6. (unordered) model_d, model_e.ipynb, model_e.after.py
"""
```

And the `1.`, `2.` is not really necessary.

I think it could be that order is conveyed vertically. And if there are several things in a single line (horizontal) it means they are unordered.

```
"""
1. model_a
2. model_a.after.py
3. model_b
4. model_c.py
5. model_c.post_hook.py, model_c.after.py
6. model_d, model_e.ipynb, model_e.after.py
"""
```

chamini2 (approved):
very cool

PR #417, opened 2022-06-22
tests: Use current_timestamp for freshness so tests work on most databases 

chamini2:


integration_tests/projects/010_source_freshness/load_freshness_table.py
```diff
@@ -14,4 +14,9 @@
 print(df)
 faldbt.write_to_source(df, "freshness_test", "freshness_table", mode="overwrite")
 
+from time import sleep
+
+# Let BigQuery cache load it
+sleep(10)
```
mederka, line 20: Could it be less than 10 seconds? This would add more minutes to our routing integration tests. Would be great to do this only if dbt-bigquery is there.

mederka (commented):
👍 

Just a minor comment

integration_tests/projects/010_source_freshness/load_freshness_table.py
chamini2: This happens only once in the tests, but yes, I think it can be less than 10 secs.

chamini2:
```suggestion
sleep(5)
```

mederka (approved):

PR #420, opened 2022-06-24
feat: Implement a DBT runner based on the DBT's Python module

isidentical:
This PR implements a second DBT runner to run DBT as a Python function. This would allow us to directly capture the `RunResults` objects (the Python object form of `run_results.json` file) without worrying about `target/` being overridden by different DBT runs. The implementation is still a work in progress.

### DIFF
--- a/src/fal/cli/dbt_runner.py
+++ b/src/fal/cli/dbt_runner.py
@@ -172,20 +172,15 @@ def _create_fal_result_file(target_path: str, run_index: int):
 # but rather use 'multiprocessing' to run it in a real system Process to
 # imitate the existing behavior of `dbt_run()` (in terms of performance).
 
-# dbt-core is currently using the spawn as its mulitprocessing context
-# so we'll mirror it.
-multiprocessing.set_start_method("spawn", force=True)
-
-
 def _dbt_run_through_python(args: List[str], target_path: str, run_index: int):
     from dbt.main import handle_and_check
 
+    run_results = exc = None
     try:
         run_results, success = handle_and_check(args)
-    except BaseException as exc:
-        run_results = None
+    except BaseException as _exc:
         return_code = getattr(exc, "code", 1)
-        traceback.print_exc()
+        exc = _exc
     else:
         return_code = 0 if success else 1
 
@@ -197,7 +192,7 @@ def _dbt_run_through_python(args: List[str], target_path: str, run_index: int):
     if run_results is not None:
         run_results.write(os.path.join(target_path, f"fal_results_{run_index}.json"))
     else:
-        raise RuntimeError("Error running dbt run")
+        raise RuntimeError("Error running dbt run") from exc
 
     return return_code
 
@@ -206,6 +201,11 @@ def dbt_run_through_python(
     args: argparse.Namespace, models_list: List[str], target_path: str, run_index: int
 ) -> DbtCliOutput:
     """Run DBT from the Python entry point in a subprocess."""
+    # dbt-core is currently using the spawn as its mulitprocessing context
+    # so we'll mirror it.
+    if multiprocessing.get_start_method() != "spawn":
+        multiprocessing.set_start_method("spawn", force=True)
+
     args = get_dbt_command_list(args, models_list, include_cli_only_args=False)
     args_as_text = " ".join(args)
     logger.info(f"Running DBT with these options: {args_as_text}")


src/fal/cli/dbt_runner.py
```diff
@@ -166,3 +160,117 @@ def _create_fal_result_file(target_path: str, run_index: int):
         shutil.copy(
             fal_run_result, os.path.join(target_path, f"fal_results_{run_index}.json")
         )
+
+
+def _process_logs(raw_logs: str):
```
turbo1912, line 165: are the logs exactly the same as it would be if we invoked the `dbt` cli command

src/fal/cli/dbt_runner.py
```diff
@@ -166,3 +160,117 @@ def _create_fal_result_file(target_path: str, run_index: int):
...
+            run_results = None
+            return_code = getattr(exc, "code", 1)
+            traceback.print_exc(exc)
+        else:
+            return_code = 0 if success else 1
+
+    logger.debug(f"dbt exited with return code {return_code}")
+
+    raw_output = output.getvalue()
+    logs, _ = _process_logs(raw_output)
```
turbo1912, line 224: are we losing streaming logs if we do this after `handle_and_check` and the patch we made to `log_manager`? I should probably try it myself..

src/fal/planner/plan.py
```diff
@@ -77,3 +77,26 @@ def plan_graph(graph: nx.DiGraph) -> nx.DiGraph:
         _reduce_subgraph(new_graph, nodes)
 
     return new_graph
+
+
+# TEMPORARY!!!
+def __test_reorder_graph(graph: nx.DiGraph) -> nx.DiGraph:
```
turbo1912, line 83: is this to be able to run integration tests? little confused about why you need this vs. writing unit tests that dont have any before or after scripts as inputs

src/fal/cli/dbt_runner.py
```diff
@@ -166,3 +160,117 @@ def _create_fal_result_file(target_path: str, run_index: int):
...
+        run_results.write(os.path.join(target_path, f"fal_results_{run_index}.json"))
+    else:
+        raise NotImplementedError("TODO: What should happen if dbt run fails and there are no run results file?")
+
+    return return_code, raw_output, logs
+
+
+def dbt_run_through_python(
+    args: argparse.Namespace, models_list: List[str], target_path: str, run_index: int
+):
```
turbo1912, line 239: return type here, at least to all public methods (still deciding if we should add a linter rule about this)

src/fal/planner/plan.py
isidentical: > is this to be able to run integration tests? little confused about why you need this vs. writing unit tests that dont have any before or after scripts as inputs

True, that is the plan for the future. I just added this to temporarily verify the execution of DBT models etc. works. Can remove it though. We currently don't have any automated tests for the parallelization, so it is for local development only.

src/fal/cli/dbt_runner.py
isidentical: Yes, I also noticed this and now it should be fixed.

src/fal/cli/dbt_runner.py
```diff
@@ -166,3 +165,76 @@ def _create_fal_result_file(target_path: str, run_index: int):
...
+# so we'll mirror it.
+multiprocessing.set_start_method("spawn", force=True)
+
+
+def _dbt_run_through_python(args: List[str], target_path: str, run_index: int):
+    from dbt.main import handle_and_check
+
+    try:
+        run_results, success = handle_and_check(args)
+    except BaseException as exc:
```
chamini2, line 185: Why not `Exception`?

src/fal/cli/dbt_runner.py
```diff
@@ -166,3 +165,76 @@ def _create_fal_result_file(target_path: str, run_index: int):
...
+        traceback.print_exc()
+    else:
+        return_code = 0 if success else 1
+
+    logger.debug(f"dbt exited with return code {return_code}")
+
+    # The 'run_results' object has a 'write()' method which is basically json.dump().
+    # We'll dump it directly to the fal results file (instead of first dumping it to
+    # run results and then copying it over).
+    if run_results is not None:
```
chamini2, line 197: Awesome

src/fal/cli/dbt_runner.py
```diff
@@ -166,3 +165,76 @@ def _create_fal_result_file(target_path: str, run_index: int):
...
+        run_results.write(os.path.join(target_path, f"fal_results_{run_index}.json"))
+    else:
+        raise RuntimeError("Error running dbt run")
```
chamini2, line 200: I wonder if instead of printing in the `except` branch and then raising here, we could just re-raise in the `except` branch with

```py
raise RuntimeError("Error running dbt run") from exc
```

To get a better stack trace

src/fal/cli/dbt_runner.py
```diff
@@ -166,3 +165,76 @@ def _create_fal_result_file(target_path: str, run_index: int):
...
+    """Run DBT from the Python entry point in a subprocess."""
+    args = get_dbt_command_list(args, models_list, include_cli_only_args=False)
+    args_as_text = " ".join(args)
+    logger.info(f"Running DBT with these options: {args_as_text}")
+
+    # We are going to use multiprocessing module to spawn a new
+    # process that will run the sub-function we have here. We
+    # don't really need a process pool, since we are only going
+    # to spawn a single process but it provides a nice abstraction
+    # over retrieving values from the process.
```
chamini2, line 217: What it does is make a "container" for running the function? That is what "abstraction" means here?

src/fal/cli/dbt_runner.py
isidentical: If there is anything that raises a `SystemExit()` (which inherits from `BaseException`), then fal will quit as well. So we are capturing that too.

src/fal/cli/dbt_runner.py
isidentical: Makes sense!

src/fal/cli/dbt_runner.py
```diff
@@ -166,3 +165,76 @@ def _create_fal_result_file(target_path: str, run_index: int):
...
+
+
+# This is the Python implementation of the `dbt_run()` function, in which
+# we directly use dbt-core as a Python library. We don't run it directly
+# but rather use 'multiprocessing' to run it in a real system Process to
+# imitate the existing behavior of `dbt_run()` (in terms of performance).
+
+# dbt-core is currently using the spawn as its mulitprocessing context
+# so we'll mirror it.
+multiprocessing.set_start_method("spawn", force=True)
```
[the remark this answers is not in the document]
isidentical: It shouldn't affect threading at any way I can imagine (since this is only for multiprocessing), but I don't think anything prevents us from shifting this inside the `dbt_run_through_python()` function (will do like that).

src/fal/cli/dbt_runner.py
isidentical: > That is what "abstraction" means here?

Normally it is hard to [retrieve the return value](https://docs.python.org/3/library/multiprocessing.html#sharing-state-between-processes) from a `multiprocessing.Process`. The `multiprocessing.Pool` however offers a nice abstraction over this problem and let you directly use the return value from a function.

turbo1912 (approved):

src/fal/cli/dbt_runner.py
chamini2: Maybe something to consider in other places later! Thanks.

mederka (approved):

src/fal/cli/dbt_runner.py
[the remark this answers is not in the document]
chamini2: The threading is from `multiprocessing` too https://github.com/fal-ai/fal/blob/bb3ded1fddff3d5eb0a3917430b8217710e9f4ae/src/fal/run_scripts.py#L5-L6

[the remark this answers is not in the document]
chamini2: I know you moved it, at this point I'm asking to learn.

chamini2 (approved):
Great!!

src/fal/cli/dbt_runner.py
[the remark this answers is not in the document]
chamini2: Cool, cool, cool. Thanks!

PR #424, opened 2022-06-27
tests: add a little delay to the post-hook script

isidentical:
The `features/flow_run_order.feature:50` test randomly fails, and from what I've gathered it seems like the timings are the problem. I am not sure whether it is definitely resolved, but I've run it ~10 times locally and it didn't fail yet.

### DIFF
--- a/integration_tests/projects/008_pure_python_models/scripts/post_hook.py
+++ b/integration_tests/projects/008_pure_python_models/scripts/post_hook.py
@@ -1,6 +1,12 @@
 import pandas as pd
 from functools import reduce
 import os
+import time
+
+# Delay just a little bit in order to make sure the file is created
+# just after the model file.
+time.sleep(0.05)
+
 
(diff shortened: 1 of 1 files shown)


isidentical:
@chamini2 If I am not missing anything, I think both of the runs you have sent seeds `008_python_models`.

image
https://github.com/fal-ai/fal/blob/7f9a6a220ab20fcd7238e87c7b35b17df0e81938/integration_tests/features/flow_run_order.feature#L51

image
https://github.com/fal-ai/fal/blob/7f9a6a220ab20fcd7238e87c7b35b17df0e81938/integration_tests/features/flow_run_order.feature#L33

mederka (approved):

isidentical:
@chamini2 btw checked the existing implementation, and don't see anything too suspicious for the post-hook runs. It is probably because the scripts are too fast I think.

PR #426, opened 2022-06-28
feat: Implement error recovery for the parallel runner

isidentical:
Depends on #423

### DIFF
--- a/src/fal/planner/executor.py
+++ b/src/fal/planner/executor.py
@@ -13,6 +13,19 @@ from fal.planner.schedule import SUCCESS, Scheduler
 from fal.planner.tasks import FalHookTask, TaskGroup, Task
 from faldbt.project import FalDbt
 
+from dbt.logger import GLOBAL_LOGGER as logger
+
+
+def _show_failed_groups(failed_groups: List[TaskGroup], skipped_groups: List[TaskGroup]) -> None:
+    failed_models = ", ".join(
+        model for group in failed_groups for model in group.task.model_ids
+    )
+    skipped_models = ", ".join(
+        model for group in skipped_groups for model in group.task.model_ids
+    )
+    logger.info(f"Failed calculating the following DBT models: {failed_models}")
+    logger.info(f"Skipped calculating the following DBT models: {skipped_models}")
+
 
 @dataclass
 class FutureGroup:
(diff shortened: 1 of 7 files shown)


src/fal/planner/schedule.py
```diff
@@ -47,7 +47,9 @@ def create_group(
 
 @dataclass
 class Scheduler:
-    groups: list[TaskGroup]
+    groups: List[TaskGroup]
+    failed_groups: List[TaskGroup] = field(default_factory=list)
+    skipped_groups: List[TaskGroup] = field(default_factory=list)
```
isidentical, line 52: Maybe we could add our version of `NodeStatus` and keep everything in a list. Like having an enum for `PENDING`, `SUCCEED`, `FAILED`, `SKIPPED`. Don't know though, this version seems very simple too. Any suggestions?

chamini2: > Maybe we could add our version of `NodeStatus` and keep everything in a list. Like having an enum for `PENDING`, `SUCCEED`, `FAILED`, `SKIPPED`

I would prefer this, tbh. then getting `failed_groups`/`skipped_groups` can be a function that just filters. But I am also OK with whatever you prefer.

chamini2 (commented):
I am kind of confused, apart from logging, what else are we achieving?

isidentical:
> I am kind of confused, apart from logging, what else are we achieving?

The error recovery was already partially implemented, but it was prevented from functioning since scheduler raised an error whenever something failed. That part is now removed, added logging for the failed/skipped nodes and tests.

chamini2 (approved):

turbo1912 (approved):

src/fal/planner/executor.py
```diff
@@ -13,6 +13,19 @@
...
+
+def _show_failed_groups(failed_groups: List[TaskGroup], skipped_groups: List[TaskGroup]) -> None:
+    failed_models = ", ".join(
+        model for group in failed_groups for model in group.task.model_ids
+    )
+    skipped_models = ", ".join(
+        model for group in skipped_groups for model in group.task.model_ids
+    )
+    logger.info(f"Failed calculating the following DBT models: {model_text}")
+    logger.info(f"Skipped calculating the following DBT models: {model_text}")
```
isidentical, line 27:
```suggestion
    logger.info(f"Failed calculating the following DBT models: {failed_models}")
    logger.info(f"Skipped calculating the following DBT models: {skipped_models}")
```

mederka (approved):

turbo1912 (approved):

src/fal/planner/schedule.py
chamini2: Are we going to change this?

PR #429, opened 2022-06-30
fix: Add lock around dbt cache manipulation

chamini2:


mederka:
Looks like a test has failed. Is there are race condition?

chamini2:
Ran it 12 times: 
image

It seems we are good.

src/faldbt/lib.py
```diff
@@ -101,6 +104,38 @@ def _get_adapter(
...
+        yield
+    except BaseException as exc:
+        _exc = exc
+
+    _lock.release()
+    logger.debug("Released {}", operationId)
+    if _exc:
+        raise RuntimeError(
+            f"Error during cache lock (operation: {operationId})"
+        ) from _exc
```
isidentical, line 131: Let's maybe just use try/finally here and propagate the exception directly (unless there is a reason to stick with `RuntimeError`, which is a behavioral change),

```py
    try:
        yield
    finally:
        _lock.release()
        logger.debug("Released {}", operationId)
```

isidentical (approved):

src/faldbt/lib.py
chamini2: But it does work. Thanks!

PR #432, opened 2022-07-01
fix: before generating sql models delete old ones

mederka:


chamini2 (approved):
Is there any way we can test this?

We had a behave step that would create files. We could use that.

isidentical (approved):

mederka:
> Is there any way we can test this?
> 
> We had a behave step that would create files. We could use that.

It looks like the behave step is script specific. I created a ticket for this and will start working on it asap:

https://linear.app/features-and-labels/issue/FEA-238/test-dbt-model-creation-tidiness

chamini2:
Can we revert this? I just realized we are losing the "We should not delete what fal did not generate" check we had with this change.

PR #443, opened 2022-07-05
fix: use correct timeout configuration based on bigquery adapter version

chamini2:


src/faldbt/lib.py
```diff
@@ -29,10 +29,10 @@
 from sqlalchemy.sql import Insert
 
 
-DBT_V1 = dbt.semver.VersionSpecifier.from_version_string("1.0.0")
-DBT_VCURRENT = dbt.version.get_installed_version()
```
isidentical, line 33: ATL: Any difference between `get_installed_version()` vs `installed`?

isidentical (approved):

src/faldbt/lib.py
chamini2: `get_installed_version()` calculates it, `installed` does not calculate it again, it is already calculated.

https://github.com/dbt-labs/dbt-core/blob/f49f28c331062c824351fe1babe4e4dd146f8d71/core/dbt/version.py#L239

PR #445, opened 2022-07-06
refactor: scheduler node statuses, only show failed/skipped nodes

isidentical:
We used to show the `failed nodes: ...` message even if everything succeeded, so this is fixed. Also refactors the node status handling.

### DIFF
--- a/src/fal/planner/tasks.py
+++ b/src/fal/planner/tasks.py
@@ -6,6 +6,7 @@ import traceback
 import uuid
 from contextlib import contextmanager
 from dataclasses import dataclass, field
+from enum import Enum, auto
 from functools import cached_property
 from typing import Iterator, List, Union, Any, Optional, Dict, Tuple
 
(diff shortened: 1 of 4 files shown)


src/fal/planner/tasks.py
```diff
@@ -24,6 +25,14 @@ def execute(self, args: argparse.Namespace, fal_dbt: FalDbt) -> int:
         raise NotImplementedError
 
 
+class GroupStatus(Enum):
+    PENDING = auto()
```
chamini2, line 29: did not know about auto

chamini2 (approved):
Pretty cool!

src/fal/planner/tasks.py
isidentical: Good for when you don't have an exact value.

isidentical:
Thanks for the reviews @chamini2!

PR #446, opened 2022-07-06
fix: ignore user warnings caused by multiprocessing

isidentical:
After each execution of `fal flow run` (with `--experimental-threads`) we were getting a user warning (like the following one)

```
10:13:28  Unable to do partial parsing because config vars, config profile, or config target have changed
Failed calculating the following DBT models: model.fal_008.broken_model
/home/isidentical/.pyenv/versions/3.9.13/lib/python3.9/multiprocessing/resource_tracker.py:229: UserWarning: resource_tracker: '/mp-zv5nn8kz': [Errno 2] No such file or directory
  warnings.warn('resource_tracker: %r: %s' % (name, e))
```

It points to the multiprocessing, which is something we only use for the `dbt_run_through_python()`. After some investigation, it seems to be a case specific to certain Python versions (3.8+?) and only due to how we use pools with threading (more information in this [bug report](https://bugs.python.org/issue38119)). Since there isn't anything we can do about it directly (as provided in the link) and since this is not a bug (or even a real resource leak) I think the easiest way go forward is just suppressing this warning.

### DIFF
--- a/src/fal/planner/executor.py
+++ b/src/fal/planner/executor.py
@@ -1,4 +1,5 @@
 import argparse
+import warnings
 from concurrent.futures import (
     FIRST_COMPLETED,
     Executor,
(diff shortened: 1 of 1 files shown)


src/fal/planner/executor.py
```diff
@@ -16,6 +17,16 @@
 from dbt.logger import GLOBAL_LOGGER as logger
 
 
+# Since we construct multiprocessing pools for each DBT run, it leaves a trace
```
chamini2, line 20: nit: What about 
 ```suggestion
# HACK: Since we construct multiprocessing pools for each DBT run, it leaves a trace
```
for easy finding later

chamini2 (approved):

PR #448, opened 2022-07-07
feat: Add support for scripts in parallel runner

isidentical:


src/fal/planner/plan.py
```diff
@@ -28,8 +29,17 @@ def from_execution_plan(
...
+            if (
+                node
+                not in execution_plan.dbt_models
+                + execution_plan.before_scripts
+                + execution_plan.after_scripts
+            ):
                 graph.remove_node(node)
+
+        if execution_plan.after_scripts:
+            logger.warn("Using after scripts is now deprecated. Please consider migrating to post-hooks.")
```
isidentical, line 41: Just to confirm: before scripts will still stick around right? @chamini2 @turbo1912

src/fal/planner/executor.py
```diff
@@ -27,20 +27,24 @@
 warnings.filterwarnings("ignore", category=UserWarning, module="multiprocessing.resource_tracker")
 
 
+def _collect_models(groups: List[TaskGroup]) -> List[str]:
+    for group in groups:
+        if not isinstance(group.task, DBTTask):
+            continue
+
+        yield from group.task.model_ids
```
chamini2, line 35:
```suggestion
        if isinstance(group.task, DBTTask):
            yield from group.task.model_ids
```

src/fal/planner/executor.py
```diff
@@ -27,20 +27,24 @@
...
-    skipped_groups = scheduler.filter_groups(GroupStatus.SKIPPED)
-    if skipped_groups:
-        skipped_models = ", ".join(
-            model for group in skipped_groups for model in group.task.model_ids
-        )
-        logger.info(f"Skipped calculating the following DBT models: {skipped_models}")
+    failed_models = _collect_models(scheduler.filter_groups(GroupStatus.FAILURE))
+    if failed_models:
+        message = ", ".join(failed_models)
+        logger.info(f"Failed calculating the following DBT models: {message}")
```
chamini2, line 42:
```suggestion
        logger.info("Failed calculating the following DBT models: {}", message)
```

src/fal/planner/plan.py
```diff
@@ -28,8 +29,17 @@ def from_execution_plan(
     ) -> FilteredGraph:
         graph = origin_graph.copy_graph()
         for node in origin_graph.graph.nodes:
-            if node not in execution_plan.dbt_models:
+            if (
+                node
+                not in execution_plan.dbt_models
+                + execution_plan.before_scripts
+                + execution_plan.after_scripts
```
chamini2, line 36: can you parenthesize this expression? I am guessing it is `not in (list+list+list)`

src/fal/planner/plan.py
```diff
@@ -28,8 +29,17 @@ def from_execution_plan(
     ) -> FilteredGraph:
         graph = origin_graph.copy_graph()
```
chamini2, line 30: Just an idea: Would we want to instead return a filtered view of the original graph? https://networkx.org/documentation/stable/reference/generated/networkx.classes.function.subgraph_view.html?highlight=filter_node

src/fal/planner/plan.py
chamini2: For now at least, this looks ok. We will have to decide later if pre-hooks will be a thing.

chamini2 (approved):
This runs the after scripts in the middle, right?

src/fal/planner/plan.py
isidentical: Ah, one catch though:

> [subgraph_view](https://networkx.org/documentation/stable/reference/generated/networkx.classes.function.subgraph_view.html?highlight=filter_node#networkx.classes.function.subgraph_view) provides a read-only view of the input graph that excludes nodes and edges based on the outcome of two filter functions filter_node and filter_edge.

We can't use it directly, since it is a read-only view (e.g., in the planner we do in-place mutations).

PR #450, opened 2022-07-07
fix: delete old generated models before fal flow run

mederka:


src/fal/cli/model_generator/model_generator.py
```diff
@@ -99,6 +103,22 @@ def _check_path_safe_to_write(sql_path: Path, py_path: Path):
...
 
+def delete_generated_models(project_dir: str):
+    model_paths = _get_model_paths(project_dir)
+    for path in model_paths:
+        for root, dirs, files in os.walk(path):
+            for file in files:
+                if file.endswith('.sql'):
+                    file_path = os.path.join(root,file)
+                    if _is_fal_generated(file_path):
+                        os.remove(file_path)
```
isidentical, line 114: I think `glob()`ing directly might simplify it a lot!
```suggestion
        for file in path.glob("**/*.sql"):
            if _is_fal_generated(file):
                file.unlink()
 ```

**Review by isidentical — APPROVED** [review]

File: src/fal/cli/model_generator/model_generator.py (hunk 1 above)
Comment (chamini2) [reply src/fal/cli/model_generator/model_generator.py:114]: Agreed with @isidentical here.

You can even abstract it as we did for Python in this same file

https://github.com/fal-ai/fal/blob/e0a01f5f0017f77ee29887d420126117153e15e6/src/fal/cli/model_generator/model_generator.py#L133-L137

Comment (chamini2) [reply src/fal/cli/model_generator/model_generator.py:114]: Maybe even generalize it:

```py
def _find_files(top_path: Path, patterns: List[str]) -> List[Path]: 
    paths_lists = (top_path.rglob(pattern) for pattern in patterns)
    paths = (p for list in paths_lists for p in list)
    return [p for p in paths if p.is_file()]
```

```py
_find_files(model_path / GENERATED_DIR, ['*.sql'])
```

src/fal/cli/model_generator/model_generator.py
```diff
@@ -99,6 +103,22 @@ def _check_path_safe_to_write(sql_path: Path, py_path: Path):
 CHECKSUM_REGEX = re.compile(r"FAL_GENERATED ([_\d\w]+)")
 
 
+def delete_generated_models(project_dir: str):
+    model_paths = _get_model_paths(project_dir)
+    for path in model_paths:
+        for root, dirs, files in os.walk(path):
```
chamini2, line 109:
```suggestion
        for root, dirs, files in os.walk(path / GENERATED_DIR):   
```

right?

chamini2 (changes requested):

mederka:
Thank you @chamini2 and @isidentical 🙏

src/fal/cli/model_generator/model_generator.py
mederka: That's a good question. Should we delete all fal generated files that might have been moved from the generated dir by mistake?

isidentical (approved):
Looks great, thanks 🚀

PR #451, opened 2022-07-07
tests: Reduce dbt versions to test on schedule and fix matrix option

chamini2:
We had too many tests scheduled: 4 Python versions and 4 dbt versions = 16 per adapter. So they were canceled because of time limits.

This brings them to 8 per adapter when scheduled and 4 per adapter for pull requests.

We can still improve by making each python version have a different schema to run adapter tests in parallel (the ones with remote data warehouse)

### DIFF
--- a/.github/workflows/test_profiles.yml
+++ b/.github/workflows/test_profiles.yml
@@ -72,15 +72,19 @@ jobs:
       list: ${{ steps.include-step.outputs.out }}
     steps:
       - id: include-step
+        shell: python
         run: |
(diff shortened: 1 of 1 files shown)


chamini2:
> We're not going to test <v1 anymore?

Only when making changes (in pull requests)

```
          if '${{ github.event_name }}' == 'pull_request':
            # Add more versions for pull requests
            OUTPUT.extend(["0.20.*", "0.21.*"])
```

mederka (approved):

PR #455, opened 2022-07-07
feat: Enable more than once write_to_model per Python Model

chamini2:


src/fal/cli/model_generator/module_check.py
```diff
@@ -34,12 +34,9 @@ def write_to_model_check(module: ast.Module):
...
     all_wtm_calls = _filter_function_calls_by_name(all_function_calls, "write_to_model")
 
-    top_level_function_calls = _find_function_calls(
-        stmt.value for stmt in module.body if isinstance(stmt, ast.Expr)
-    )
     assert (
-        len(all_wtm_calls) == 1 and all_wtm_calls[0] in top_level_function_calls
-    ), "There must be a single write_to_model call and it must be a top-level expression of the Python Model script"
+        len(all_wtm_calls) > 0
+    ), "There must be at least a write_to_model call in the Python Model"
```
isidentical, line 39:
```suggestion
    ), "There must be at least one write_to_model call in the Python Model"
```

mederka (approved):
This is very tidy, @chamini2. Maybe I don't quite understand the point of this though. We are writing to the same model right? Why would we want to do it multiple times?

chamini2:
A user could do an initial overwrite and then appends in the same script and we had agreed it was too strict of a check.

burkaygur (approved):

PR #460, opened 2022-07-12
feat: Make before/after scripts happen right before/after models

chamini2:
Before _shuffling_ (I can change that name) them
before

And afterward
after

<details>
closes FEA-38
</details>

### DIFF
--- a/integration_tests/projects/008_pure_python_models/scripts/before.py
+++ b/integration_tests/projects/008_pure_python_models/scripts/before.py
@@ -1,0 +1,3 @@
+from _fal_testing import create_post_hook_artifact
+
+create_post_hook_artifact(context, "before")
--- a/src/fal/cli/flow_runner.py
+++ b/src/fal/cli/flow_runner.py
@@ -50,7 +50,7 @@ def run_threaded(
         OriginGraph,
         FilteredGraph,
         PlannedGraph,
-        ShuffledGraph,
+        ScriptConnectedGraph,
     )
     from fal.planner.schedule import schedule_graph
     from fal.planner.executor import parallel_executor
@@ -64,9 +64,9 @@ def run_threaded(
     filtered_graph = FilteredGraph.from_execution_plan(
         origin_graph, execution_plan=execution_plan
     )
-    shuffled_graph = ShuffledGraph.from_filtered_graph(filtered_graph)
-    planned_graph = PlannedGraph.from_shuffled_graph(
-        shuffled_graph, enable_chunking=False
+    connected_graph = ScriptConnectedGraph.from_filtered_graph(filtered_graph)
+    planned_graph = PlannedGraph.from_script_connected_graph(
+        connected_graph, enable_chunking=False
     )
     scheduler = schedule_graph(planned_graph.graph, node_graph)
     parallel_executor(
--- a/src/fal/cli/flow_runner.py
+++ b/src/fal/cli/flow_runner.py
@@ -50,6 +50,7 @@ def run_threaded(
         OriginGraph,
         FilteredGraph,
         PlannedGraph,
+        ShuffledGraph,
     )
     from fal.planner.schedule import schedule_graph
     from fal.planner.executor import parallel_executor
@@ -63,8 +64,9 @@ def run_threaded(
     filtered_graph = FilteredGraph.from_execution_plan(
         origin_graph, execution_plan=execution_plan
     )
-    planned_graph = PlannedGraph.from_filtered_graph(
-        filtered_graph, enable_chunking=False
+    shuffled_graph = ShuffledGraph.from_filtered_graph(filtered_graph)
+    planned_graph = PlannedGraph.from_shuffled_graph(
+        shuffled_graph, enable_chunking=False
     )
     scheduler = schedule_graph(planned_graph.graph, node_graph)
     parallel_executor(
--- a/src/fal/planner/plan.py
+++ b/src/fal/planner/plan.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from typing import Iterator, List
+from typing import Iterator, List, Set
+import re
 
 import networkx as nx
 from fal.node_graph import NodeKind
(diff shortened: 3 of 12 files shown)


src/fal/cli/flow_runner.py
```diff
@@ -63,8 +64,9 @@ def run_threaded(
     filtered_graph = FilteredGraph.from_execution_plan(
         origin_graph, execution_plan=execution_plan
     )
-    planned_graph = PlannedGraph.from_filtered_graph(
-        filtered_graph, enable_chunking=False
+    shuffled_graph = ShuffledGraph.from_filtered_graph(filtered_graph)
+    planned_graph = PlannedGraph.from_shuffled_graph(
+        shuffled_graph, enable_chunking=False
```
mederka, line 69: ATL: What's happening here?

src/fal/planner/plan.py
```diff
@@ -14,7 +15,32 @@ class OriginGraph:
...
+        return self.graph.copy()  # type: ignore
+
+    def _plot(self, graph=None):
+        """
+        For development and debugging purposes
+        """
+        if not graph:
+            graph = self.graph
+
+        import matplotlib.pyplot as plt
```
mederka, line 27: Is this a dev dependency or does come automatically with networkx?

src/fal/planner/plan.py
```diff
@@ -41,16 +67,74 @@ def from_execution_plan(
...
+                # Keep the original `node` to `succ` edge and add a new one from `script`
+                self.graph.add_edges_from([(script, succ) for script in after_scripts])
+
+                # And add an edge between all node after scripts to the succ before scripts
+                succ_predecessors = old_graph.predecessors(succ)
+                # TODO: is_before_script() later on!!
+                succ_before_scripts = _pattern_matching(
+                    ShuffledGraph.before_pattern, succ_predecessors
+                )
+                for before_script in succ_before_scripts:
```
mederka, line 111: Maybe we can refactor this method to make it more readable. Otherwise we're in a for loop inside a for loop inside a for loop and it's a little tough to wrap your head around it.

mederka (commented):
Looks great, @chamini2! I have minor comments. Also curious about @isidentical's thoughts on `ShuffledGraph`.

src/fal/planner/plan.py
chamini2: Dev dependency. I will add it to pyproject.toml

src/fal/cli/flow_runner.py
chamini2: `PlannedGraph.from_shuffled_graph` does nothing other than wrap the graph when `enable_chunking=False`

src/fal/planner/plan.py
chamini2: done

integration_tests/projects/008_pure_python_models/scripts/before.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+model_name = context.current_model.name
+
+output = model_name
+
+path = os.path.join(os.environ["temp_dir"], model_name + ".before.txt")
+with open(path, "w") as file:
+    file.write(output)
```
isidentical, line 9: Let's start using the `_fal_testing.create_*` stuff (with the `prompt="before"`)
https://github.com/fal-ai/fal/blob/4373278c33c8b34214a434d127cb82d2de1b3a96/tests/_fal_testing/utils.py#L22-L23

src/fal/cli/flow_runner.py
```diff
@@ -50,6 +50,7 @@ def run_threaded(
         OriginGraph,
         FilteredGraph,
         PlannedGraph,
+        ShuffledGraph,
```
isidentical, line 53: I'd go with `Reordered` or `Connected` instead of `Shuffled`, no strong opinions though (shuffling is generally used in the context of the arrangement being random).

src/fal/planner/plan.py
```diff
@@ -41,16 +67,73 @@ def from_execution_plan(
 
 
 @dataclass
-class PlannedGraph(OriginGraph):
+class ShuffledGraph(OriginGraph):
     graph: nx.DiGraph
+    after_pattern = re.compile("^script\\..+\\.AFTER\\..+\\.(py|ipynb)")
+    before_pattern = re.compile("^script\\..+\\.BEFORE\\..+\\.(py|ipynb)")
```
isidentical, line 73: ATL: why not use `fal.cli.selectors._is_before_script`?
https://github.com/fal-ai/fal/blob/4373278c33c8b34214a434d127cb82d2de1b3a96/src/fal/cli/selectors.py#L287-L296

src/fal/planner/plan.py
```diff
@@ -41,16 +67,73 @@ def from_execution_plan(
...
+        def _add_edges_from_to(from_nodes: Set[str], to_nodes: Set[str]):
+            self.graph.add_edges_from(
+                (from_n, to_n) for from_n in from_nodes for to_n in to_nodes
+            )
+
+        old_graph = self.copy_graph()
+        node: str
+        for node in old_graph.nodes:
+
+            # TODO: is_after_script() later on!!
```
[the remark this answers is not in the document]
isidentical: I thought it would be easier for us to not to copy over the same matcher regex.

src/fal/planner/plan.py
```diff
@@ -41,16 +67,73 @@ def from_execution_plan(
...
+            after_scripts, other_succs = _pattern_matching(
+                ShuffledGraph.after_pattern, set(old_graph.successors(node))
+            )
+            # Keep the original node to succs edges and add a new one from the script to succs
+            _add_edges_from_to(after_scripts, other_succs)
```
isidentical, line 106: You can also remove the edges from `node` to `other_succs` (for better calculation in scheduler's dependency management code)

isidentical:
Overall LGTM, a few minor questions/details @chamini2.

src/fal/cli/flow_runner.py
chamini2: Agreed, I will change it!

src/fal/planner/plan.py
chamini2: I had forgotten about this. I'll use it

src/fal/planner/plan.py
chamini2: Thinking...

```mermaid
flowchart LR
  a --> b & a_after1 & a_after2;
```

with your proposal, it becomes

```mermaid
flowchart LR
  a --> a_after1 & a_after2;
  a_after1 & a_after2 --> b;
```

current solution does

```mermaid
flowchart LR
  a --> a_after1 & a_after2 & b;
  a_after1 & a_after2 --> b;
```

*****

Regarding selection:

`--select a+1`: 

```mermaid
flowchart LR
  a --> a_after1 & a_after2 & b;
  a_after1 & a_after2 --> b;
  classDef running fill:green;
  class a,a_after1,a_after2,b running;
```

vs


```mermaid
flowchart LR
  a --> a_after1 & a_after2;
  a_after1 & a_after2 --> b;
  classDef running fill:green;
  class a,a_after1,a_after2 running;
```

isidentical: But doesn't the selection happen before we re-order the graph @chamini2? We calculate the execution plan according to the *original graph* (without re-ordering) and then once we get the `FilteredGraph` we can safely re-order everything without worrying about how it would change the behavior for the users.

src/fal/planner/plan.py
```diff
@@ -41,16 +66,70 @@ def from_execution_plan(
...
+        def get_after_scripts(graph: nx.DiGraph, node: str):
+            return _pattern_matching(_is_after_script, set(graph.successors(node)))
+
+        def _add_edges_from_to(from_nodes: Set[str], to_nodes: Set[str]):
+            self.graph.add_edges_from(
+                (from_n, to_n) for from_n in from_nodes for to_n in to_nodes
+            )
+
+        old_graph = self.copy_graph()
+        node: str
```
isidentical, line 101: We probably don't need this.
```suggestion
```

isidentical (approved):
Thanks a lot @chamini2, LGTM (we can defer the dropping node connections for a future PR, since it is theoretically an optimization).

src/fal/planner/plan.py
chamini2: If I comment it:
image

PR #464, opened 2022-07-15
chore: Remove support for dbt 0.X

chamini2:
Merge when releasing 0.4

wait for #470 
<details>
closes FEA-256
</details>


isidentical (approved):
Should we also add a telemetry call about what version people use?

PR #466, opened 2022-07-18
feat: Offer type annotations for exposed functions.

isidentical:
image
image
image
image


src/fal/context.py
```diff
@@ -0,0 +1,107 @@
...
+# don't manually alter the contents.
+
+from __future__ import annotations
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import pandas as pd
+    from faldbt.project import DbtModel, DbtTest, Feature
+    from fal.fal_script import Context
+    from typing import Optional, Dict, List, Protocol, Any
```
isidentical, line 11: The reason we use protocols here is the `Callable[]` syntax can't really represent some stuff (like defaults and doc-strings) https://mypy.readthedocs.io/en/stable/protocols.html#callback-protocols

chamini2 (approved):
This is PRETTY COOL!

integration_tests/projects/008_pure_python_models/scripts/post_hook.py
```diff
@@ -2,6 +2,7 @@
 from functools import reduce
 import os
 import time
+from fal.context import *
```
turbo1912, line 5: and everything still works if you dont import this right?

turbo1912 (approved):

integration_tests/projects/008_pure_python_models/scripts/post_hook.py
isidentical: yep!

PR #471, opened 2022-07-21
feat: Implement pre-hooks

isidentical:
Here is a simple flow chart:

```mermaid
stateDiagram
    state "Execute the main task" as task
    state "Run the pre-hooks" as pre_hook
    state "Run the post-hooks" as post_hook
 
    [*] --> pre_hook
    pre_hook --> task: on success
    pre_hook --> post_hook: on failure

    task --> post_hook: both on success and failure
    post_hook --> [*]
```

### DIFF
--- a/src/faldbt/project.py
+++ b/src/faldbt/project.py
@@ -2,6 +2,7 @@ from collections import defaultdict
 import os.path
 import re
 from dataclasses import dataclass, field
+from functools import partialmethod
 from typing import Dict, Iterable, List, Any, Optional, Tuple, Sequence, Union
 from pathlib import Path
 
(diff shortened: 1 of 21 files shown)


isidentical:
Are we planning `pre-hook` support for `fal run`? (not `flow run`)

chamini2:
> Are we planning `pre-hook` support for `fal run`? (not `flow run`)

Just if `post-hook` is supported. But it can be a separate PR/

src/faldbt/project.py
```diff
@@ -172,18 +173,21 @@ def __hash__(self) -> int:
     def get_depends_on_nodes(self) -> List[str]:
         return self.node.depends_on_nodes
 
-    def get_post_hook_paths(self, keyword: str = "fal") -> List[str]:
+    def _get_hook_paths(self, keyword: str) -> List[str]:
```
chamini2, line 176: We need the `keyword="fal"` param because we have the `--keyword` flag. Can you re-add it and add a new param for the `hook_keyword`?

isidentical: Good catch! I search through the code base for `get_post_hook_paths` and none of the uses was passing `keyword` so I thought it was obsolete but I think it is a bug right?

https://github.com/fal-ai/fal/blob/d2a2022a292989a83d9ea767b8b3f9e00a041605/src/fal/cli/fal_runner.py#L88

E.g. should the use case above pass `faldbt.keyword`?

chamini2: probably!

chamini2 (approved):

PR #472, opened 2022-07-22
fix: Handle partially-applied write_to_model before scripts in typing module

chamini2:
image

### DIFF
--- a/src/fal/typing.py
+++ b/src/fal/typing.py
@@ -60,8 +60,8 @@ if TYPE_CHECKING:
             data: pd.DataFrame,
             target_source_name: str,
             target_table_name: str,
-            dtype: Any = None,
             *,
+            dtype: Any = None,
             mode: str = "append",
         ):
             """
--- a/src/fal/typing.py
+++ b/src/fal/typing.py
@@ -68,6 +68,17 @@ if TYPE_CHECKING:
             Write a pandas.DataFrame to a dbt source automagically.
             """
 
+    class _Write_To_Firestore(Protocol):
+        def __call__(self, df: pd.DataFrame, collection: str, key_column: str):
+            """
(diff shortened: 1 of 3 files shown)


isidentical:
@chamini2 can you share some context on what is the issue (I am not really familiar with the implementation of `write_to_model`).

chamini2:
https://github.com/fal-ai/fal/blob/main/src/fal/fal_script.py#L115-L119

This happens before the script starts, we don't really send the FalDbt version down to the script.

isidentical:
Ah, that's interesting thanks!!

isidentical:
@chamini2 since this is a very specific case and handling of this inside the function is very convoluted as you noted, would it make sense to statically embed this just like `context`? Like this https://gist.github.com/isidentical/d337d0e2e6597bc3ef3ca2a097150a81

chamini2:
> @chamini2 since this is a very specific case and handling of this inside the function is very convoluted as you noted, would it make sense to statically embed this just like `context`? Like this https://gist.github.com/isidentical/d337d0e2e6597bc3ef3ca2a097150a81
> 
> I am fine with this version too (it preserves the doc-strings and other changes in the future, so we might just bear the complexity), the example above is just a suggestion to your todo comment.

Let's go this way, but your suggestion is not bad to be honest. We can do that later if we feel it's not worth keeping.

chamini2:
> @chamini2 since this is a very specific case and handling of this inside the function is very convoluted as you noted, would it make sense to statically embed this just like `context`? Like this https://gist.github.com/isidentical/d337d0e2e6597bc3ef3ca2a097150a81
> 
> I am fine with this version too (it preserves the doc-strings and other changes in the future, so we might just bear the complexity), the example above is just a suggestion to your todo comment.

I liked this. Done!

isidentical (approved):

PR #474, opened 2022-07-25
tests: add support for pre-hooks on the DAG verifier

isidentical:


chamini2 (approved):

integration_tests/features/steps/fal_steps.py
```diff
@@ -219,43 +219,67 @@ def _verify_node_order(context):
...
-            _script_filename(post_hook)
-            for post_hook in data.get("post_hook", [])
-            if _script_filename(post_hook) in ordered_nodes
-        ]
-
+        for container, hook_type in [
+            (pre_hooks, "pre_hook"),
+            (post_hooks, "post_hook"),
+        ]:
+            container[name] = [
```
chamini2, line 246: Wait, aren't you overwriting the value here?

isidentical: not overwriting, but setting. We only have 1 write against `container[name]`.

PR #483, opened 2022-07-26
feat: optimize write and read for snowflake

chamini2:
https://github.com/snowflakedb/snowflake-connector-python/issues/544 manually resolved

Merge #487 before and use it's introduced artifact creation for tests.

<details>
Affected

- bigquery
- snowflake
</details>

### DIFF
--- a/src/faldbt/lib.py
+++ b/src/faldbt/lib.py
@@ -1,5 +1,6 @@
 # NOTE: INSPIRED IN https://github.com/dbt-labs/dbt-core/blob/43edc887f97e359b02b6317a9f91898d3d66652b/core/dbt/lib.py
 from contextlib import contextmanager
+import datetime
 import six
 from enum import Enum
 from dataclasses import dataclass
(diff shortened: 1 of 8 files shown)


tests/_fal_testing/utils.py
```diff
@@ -1,29 +1,31 @@
...
+    return output
 
 
-def create_script_artifact(context, prompt):
-    create_artifact(context, f".{prompt}.txt")
+def create_model_artifact(context, **kwargs):
+    create_artifact(context, ".txt", **kwargs)
 
 
-create_before_script_artifact = partial(create_script_artifact, prompt="before")
```
isidentical, line 26: Why are we removing these? (I also decided to remove them in the structured hooks PR, but not sure why their removal is needed here)

chamini2: not needed, more of an opinion. It came from needing to replace the content created inside the artifact.

isidentical: Wdyt about the approach in the structed hooks PR (`create_dynamic_artifact`)?

chamini2: I think that works. I can remove these changes.

isidentical: Thanks!

src/faldbt/lib.py
```diff
@@ -688,3 +688,87 @@ def _bigquery_write_relation(
...
+                schema=schema,
+                auto_create_table=True,
+                quote_identifiers=False,
+            )
+            if not success:
+                # In case the failure does not raise by itself
+                # I have not been able to reproduce such a case
+                from dbt.exceptions import DatabaseException
+
+                raise DatabaseException(output)
```
isidentical, line 771: Nice!

isidentical (approved):

PR #487, opened 2022-07-28
feat: structured hooks for passing arguments

isidentical:
Depends on #485. This PR adds support for structured hooks, hooks where you can pass arguments. E.g,
```
models:
  - name: model_a
    meta:
      fal:
        pre-hook:
          - path: add.py
            with:
              left: 2
              right: 3
```

`path` is the keyword for local hooks, and in the future, we might introduce `url` for the fal packages.


integration_tests/projects/013_structured_hooks/models/schema.yml
```diff
@@ -0,0 +1,21 @@
...
+              left: 2
+              right: 3
+
+        post-hook:
+          - post_hook_1.py
+          - path: post_hook_2.py
+          - path: sub.py
+            with:
+              left: 5
+              right: 2
```
chamini2, line 21: can we test list/dict arguments?

src/fal/fal_script.py
```diff
@@ -16,6 +16,28 @@
 from dbt.contracts.graph.parsed import ColumnInfo
 
 
+class Hook:
+    ...
```
chamini2, line 20: TIL

src/fal/fal_script.py
```diff
@@ -16,6 +16,28 @@
...
+
+
+def create_hook(raw_hook: Any) -> Hook:
+    if isinstance(raw_hook, str):
+        return LocalHook(raw_hook)
+    elif isinstance(raw_hook, dict):
+        if "path" in raw_hook:
+            return LocalHook(raw_hook["path"], raw_hook.get("with", {}))
+        else:
+            raise ValueError(f"A structured hook must specify the 'path'.")
```
chamini2, line 36: NIT
```suggestion
            raise ValueError(f"A hook must specify path.")
```

src/fal/fal_script.py
```diff
@@ -53,30 +75,55 @@ def __init__(self, config: RuntimeConfig):
...
     current_model: Union[CurrentModel, None]
     config: ContextConfig
+    _arguments: Optional[Dict[str, Any]] = field(repr=False, default=None)
+
+    @property
+    def arguments(self) -> Dict[str, Any]:
+        if self._arguments is None:
+            raise ValueError(
+                "'context.arguments' is only accessible from hooks, "
+                "not from scripts/models"
```
chamini2, line 85: Why do you think is valuable to make this distinction?

src/fal/fal_script.py
```diff
@@ -156,7 +203,9 @@ def is_model(self):
     def model_name(self):
         return "<GLOBAL>" if self.is_global else self.model.name  # type: ignore
 
-    def _build_script_context(self):
+    def _build_script_context(
+        self, script_arguments: Optional[Dict[str, Any]] = None
```
chamini2, line 207: we do not use `script_arguments`

src/fal/planner/tasks.py
```diff
@@ -207,17 +208,23 @@ def execute(self, args: argparse.Namespace, fal_dbt: FalDbt) -> int:
 
     @classmethod
     def from_fal_script(cls, script: FalScript):
-        return cls(script.path, script.model, script.is_hook)
+        return cls(script.path, script.model, script.hook_arguments, script.is_hook)
```
chamini2, line 211: NIT: maybe change to multi-line?

src/faldbt/project.py
```diff
@@ -43,6 +53,9 @@
 
 from fal.telemetry import telemetry
 
+if TYPE_CHECKING:
```
chamini2, line 56: this avoids a circular dependency?

src/faldbt/project.py
```diff
@@ -173,21 +186,26 @@ def __hash__(self) -> int:
     def get_depends_on_nodes(self) -> List[str]:
         return self.node.depends_on_nodes
 
-    def _get_hook_paths(self, hook_type: str, keyword: str = "fal") -> List[str]:
+    def _get_hooks(self, hook_type: str, keyword: str = "fal") -> List["Hook"]:
```
chamini2, line 189: ATL: if it is imported, why not `List[Hook]`?

src/fal/fal_script.py
isidentical: > Why do you think is valuable to make this distinction?

Not a distinction per se, but this is specifying the remaining types of code that can access `context` (`pre-hook`/`post-hook` can access it, `scripts` and `python models` can't.)

src/faldbt/project.py
isidentical: yes

src/faldbt/project.py
isidentical: It is not imported, the import for this is guarded by `TYPE_CHECKING` (to prevent circular dependencies) so we can't directly reference the name but rather use the stringified version.

integration_tests/projects/013_structured_hooks/models/schema.yml
```diff
@@ -0,0 +1,31 @@
...
+          - path: types.py
+            with:
+              number: 5
+              text: "type"
+              sequence:
+                - 1
+                - 2
+                - 3
+              mapping:
+                key: value
```
chamini2, line 31: awesome

integration_tests/projects/013_structured_hooks/scripts/types.py
```diff
@@ -0,0 +1,8 @@
+from fal.typing import *
+from _fal_testing import create_dynamic_artifact
+
+create_dynamic_artifact(
+    context,
+    "Types: "
+    + ", ".join(f"{key}={type(value).__name__}" for key, value in context.arguments.items()),
```
chamini2, line 7: can we also print `value`?

```suggestion
    + ", ".join(f"{key}={value} ({type(value).__name__})" for key, value in context.arguments.items()),
```

src/fal/fal_script.py
```diff
@@ -16,6 +16,28 @@
...
+
+
+def create_hook(raw_hook: Any) -> Hook:
+    if isinstance(raw_hook, str):
+        return LocalHook(raw_hook)
+    elif isinstance(raw_hook, dict):
+        if "path" in raw_hook:
+            return LocalHook(raw_hook["path"], raw_hook.get("with", {}))
+        else:
+            raise ValueError(f"A hook must specify path.")
```
chamini2, line 36: Could we add tests for this? and maybe for trying to use dict formatting for after/before-scripts.

chamini2 (approved):
test ideas only. Great work!

integration_tests/projects/013_structured_hooks/scripts/types.py
isidentical: sure!

src/fal/fal_script.py
isidentical: The PR for this seems to be much bigger than anticipated, so I'll do it separately.

isidentical: RE: #494

PR #489, opened 2022-07-29
feat: implement isolated hooks

isidentical:
This PR implements "isolated hooks" (e.g., hooks which run in a different environment than the fal itself). Current implementation:

- Only uses virtual envs, so only requirements that can be installed with pip is available
- Virtual envs that are created for an environment set is cached. So in the second run for the same hook (or a different hook that has the same exact dependencies) it will be very fast.
- Each environment contains `fal` + `dbt-core` and all the dbt adapters you have locally (we scan all the installed python packages you have, and install the same version of every `dbt-*` adapter)

### Example
```yaml
  - name: model_b
    meta:
      fal:
        pre-hook:
         # This is a regular hook, that runs on the same environment as the fal itself
          - path: local_hook.py

          # This is an isolated hook, which runs in an environment called 'funny'. These
          # environments are defined inside fal_project.yaml, and uses the same format
          # as the "fal runtime" (so they are interchangeable).
          - path: funny_hook.py
            environment: funny
           
          # You can specify arguments just like you could for local hooks, and since
          # arguments are yaml values the serialization is very easy!
          - path: check_imports.py
            environment: not-funny
            with:
              import: pyjokes
              expected_success: false
```

Here is the environment spec:
```yaml
environments:
  # An environment without a 'requirements' key is considered a bare virtual
  # environment where you want to run your code without the affect of other
  # threads. This one only includes fal + dbt-core + dbt-* adapters
  - name: not-funny
  
  # This one defines some requirements, so they are also installed in addition to the
  # defined requirements above.
  - name: funny
    requirements:
      - pyjokes==0.6.0  
```

### DIFF
--- a/integration_tests/projects/013_structured_hooks/models/schema.yml
+++ b/integration_tests/projects/013_structured_hooks/models/schema.yml
@@ -29,3 +29,17 @@ models:
                 - 3
               mapping:
                 key: value
+
+  - name: model_b
+    meta:
+      fal:
+        pre-hook:
+          - path: local_hook.py
+          - path: funny_hook.py
+            environment: funny
+        post-hook:
+          - path: check_imports.py
+            environment: not-funny
+            with:
+              import: pyjokes
+              expected_success: false
--- a/src/fal/packages/_run_hook.py
+++ b/src/fal/packages/_run_hook.py
@@ -1,0 +1,54 @@
+from re import L
+from logbook import StreamHandler
+import json
+import sys
+from argparse import ArgumentParser, Namespace
+from pathlib import Path
+from typing import Any, Dict
+
+
+def reconfigure_handlers():
+    from logbook import INFO, StreamHandler
+    from dbt.logger import GLOBAL_LOGGER as logger
+
+    # Reconfigure DBT's log handlers to output to
+    # stdout so we can stream it back to the parent
+    # process.
+    logger.disabled = False
+    logger.handlers.append(StreamHandler(sys.stdout))
+    logger.level = INFO
+
+def run_fal_hook(
+    path: str,
+    bound_model_name: str,
+    fal_dbt_config: Dict[str, Any],
+    arguments: Dict[str, Any],
+    run_index: int,
+) -> None:
+    from fal.node_graph import DbtModelNode, NodeGraph
+    from fal.planner.tasks import FalLocalHookTask
+    from faldbt.project import FalDbt
+
+    fal_dbt = FalDbt(**fal_dbt_config)
+    node_graph = NodeGraph.from_fal_dbt(fal_dbt)
+    flow_node = node_graph.get_node(bound_model_name)
+    assert isinstance(flow_node, DbtModelNode)
+
+    task = FalLocalHookTask(path, flow_node.model, arguments)
+    task._run_index = run_index
+
+    dummy_namespace = Namespace()
+    return task.execute(dummy_namespace, fal_dbt)
+
+
+def main() -> None:
+    parser = ArgumentParser()
+    parser.add_argument("data", type=str)
+
+    options = parser.parse_args()
+    exit_code = run_fal_hook(**json.loads(options.data))
+    sys.exit(exit_code)
+
+
+if __name__ == "__main__":
+    main()
--- a/src/fal/packages/_run_hook.py
+++ b/src/fal/packages/_run_hook.py
@@ -1,22 +1,11 @@
-from re import L
-from logbook import StreamHandler
 import json
 import sys
 from argparse import ArgumentParser, Namespace
 from pathlib import Path
 from typing import Any, Dict
+from fal.cli.cli import reconfigure_logging
+from dbt.logger import log_manager
 
-
-def reconfigure_handlers():
-    from logbook import INFO, StreamHandler
-    from dbt.logger import GLOBAL_LOGGER as logger
-
-    # Reconfigure DBT's log handlers to output to
-    # stdout so we can stream it back to the parent
-    # process.
-    logger.disabled = False
-    logger.handlers.append(StreamHandler(sys.stdout))
-    logger.level = INFO
 
 def run_fal_hook(
     path: str,
@@ -24,6 +13,7 @@ def run_fal_hook(
     fal_dbt_config: Dict[str, Any],
     arguments: Dict[str, Any],
     run_index: int,
+    disable_logging: bool,
 ) -> None:
     from fal.node_graph import DbtModelNode, NodeGraph
     from fal.planner.tasks import FalLocalHookTask
@@ -38,7 +28,9 @@ def run_fal_hook(
     task._run_index = run_index
 
     dummy_namespace = Namespace()
-    return task.execute(dummy_namespace, fal_dbt)
+    reconfigure_logging(disable_logging=disable_logging)
+    with log_manager.applicationbound():
+        return task.execute(dummy_namespace, fal_dbt)
 
 
 def main() -> None:
--- a/src/fal/packages/environment.py
+++ b/src/fal/packages/environment.py
@@ -1,0 +1,157 @@
+from __future__ import annotations
+
+import hashlib
+import json
+import shutil
+import subprocess
+import threading
+from collections import defaultdict
+from contextlib import contextmanager
+from dataclasses import dataclass, field
+from functools import cached_property, partial
+from pathlib import Path
+from typing import (
+    Any,
+    Callable,
+    ClassVar,
+    ContextManager,
+    DefaultDict,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    Tuple,
+)
+
+from platformdirs import user_cache_dir
+
+import fal.packages._run_hook as _hook_runner_module
+from faldbt.project import FalDbt
+from fal.planner.tasks import SUCCESS, FAILURE
+
+_BASE_CACHE_DIR = Path(user_cache_dir("fal", "fal"))
+_BASE_CACHE_DIR.mkdir(exist_ok=True)
+
+_BASE_VENV_DIR = _BASE_CACHE_DIR / "venvs"
+_BASE_VENV_DIR.mkdir(exist_ok=True)
+
+# TODO: for dev versions, we need a better solution.
+_IS_DEV_VERSION = True
+
+
+def _get_default_requirements() -> Iterator[Tuple[str, Optional[str]]]:
+    import importlib_metadata
+
+    # TODO: for dev versions, we need a better solution.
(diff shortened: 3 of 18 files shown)


isidentical:
Requires #487 and #491

integration_tests/projects/013_structured_hooks/models/schema.yml
```diff
@@ -29,3 +29,17 @@ models:
...
+      fal:
+        pre-hook:
+          - path: local_hook.py
+          - path: funny_hook.py
+            environment: funny
+        post-hook:
+          - path: check_imports.py
+            environment: not-funny
+            with:
+              import: pyjokes
```
mederka, line 44: atl: what's the benefit of specifying imports here?

mederka (approved):
This is great!

src/fal/packages/environment.py
```diff
@@ -0,0 +1,157 @@
...
+    def _create_venv(self) -> Path:
+        from virtualenv import cli_run
+
+        path = _BASE_VENV_DIR / self._key
+        with self._VENV_LOCKS[self._key]:
+            with _clear_on_fail(path):
+                if path.exists():
+                    return path
+
+                print(f"Creating virtual environment at {path}.")
```
chamini2, line 112: ATL: why are we printing instead of using a logger? Is it not available in this context?

src/fal/packages/environment.py
```diff
@@ -0,0 +1,157 @@
...
+        process = subprocess.Popen(
+            [
+                python_path,
+                _hook_runner_module.__file__,
+                data,
+            ],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            text=True,
+        )
```
chamini2, line 150: ATL: why do we do a subprocess instead of running in this same process? it seems we are already in a thread, right?

src/faldbt/project.py
```diff
@@ -436,6 +433,16 @@ def __init__(
...
 
+    def _serialize(self) -> Dict[str, Any]:
+        # All the information required for re-creating the project
+        return {
+            "project_dir": self.project_dir,
+            "profiles_dir": self.profiles_dir,
+            "threads": self.threads,
+            "keyword": self.keyword,
+            "profile_target": self._profile_target,
+        }
```
chamini2, line 444: this is surprisingly small

chamini2 (commented):
Please don't merge until we discuss these. Thanks for very interesting PR!

src/fal/packages/_run_hook.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def reconfigure_handlers():
+    from logbook import INFO, StreamHandler
+    from dbt.logger import GLOBAL_LOGGER as logger
+
+    # Reconfigure DBT's log handlers to output to
+    # stdout so we can stream it back to the parent
+    # process.
+    logger.disabled = False
```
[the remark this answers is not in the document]
chamini2: https://github.com/fal-ai/fal/blob/c70a2f4b37621accbd256d85de5e5181e866cf16/src/fal/cli/cli.py#L33-L38

src/faldbt/project.py
isidentical: Not sure if there is any missing one, but these are all I found.

integration_tests/projects/013_structured_hooks/models/schema.yml
isidentical: Just to play with the combinations of arguments to hooks, and writing a hook that can actually be a test of itself.

src/fal/packages/_run_hook.py
```diff
@@ -0,0 +1,54 @@
+from re import L
```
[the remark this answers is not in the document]
isidentical: nop!

src/fal/packages/environment.py
```diff
@@ -0,0 +1,157 @@
...
+    # *at the same time* it might fail (the behavior is undefined)·
+    #
+    # We might think about introducing a file-lock, which would
+    # allow this scenerio in the future.
+    _VENV_LOCKS: ClassVar[DefaultDict[threading.Lock]] = defaultdict(threading.Lock)
+
+    def __post_init__(self) -> None:
+        self.requirements.extend(
+            f"{key}=={value}" if value else key
+            for key, value in _get_default_requirements()
```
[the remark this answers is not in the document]
chamini2: OK, I would say let's ignore fal or dbt versions and log a warning for them

src/fal/packages/environment.py
```diff
@@ -0,0 +1,157 @@
...
+    # TODO: for dev versions, we need a better solution.
+    if _IS_DEV_VERSION:
+        yield "git+https://github.com/fal-ai/fal@batuhan/fea-318-use-the-environment-option-for-local", None
+    else:
+        yield "fal", importlib_metadata.version("fal")
+    yield "dbt-core", importlib_metadata.version("dbt-core")
+
+    for dynamic_dependency in importlib_metadata.distributions():
+        if dynamic_dependency.name.startswith("dbt-"):
+            yield dynamic_dependency.name, dynamic_dependency.version
```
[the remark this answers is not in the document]
chamini2: > I wish it was part of the public API though, we could have just used it instead of copying this complex snippet.

I know!

src/fal/packages/environment.py
chamini2: Of course! This makes a lot of sense. Very cool experiments, though.
I agree that it is very complex without really adding substantial value. Not worth doing.

Thanks for the explanation!

src/fal/packages/_run_hook.py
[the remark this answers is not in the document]
chamini2: > ATIL: what is new dbt logging?

https://github.com/dbt-labs/dbt-core/blob/v1.0.0/core/dbt/events/README.md#adapter-maintainers

We had not migrated because 0.X dbt did not have the new logging available. I will get it done soon!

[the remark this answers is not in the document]
chamini2: I think the initial idea of just adding a `TODO` comment in here to remember to take a look is enough. To not block this PR at all.

src/fal/packages/environment.py
```diff
@@ -0,0 +1,164 @@
...
+        self.requirements.extend(
+            f"{key}=={value}" if value else key
+            for key, value in _get_default_requirements()
+        )
+
+    @property
+    def _key(self) -> str:
+        crypt = hashlib.sha256()
+        crypt.update(" ".join(self.requirements).encode())
+        return crypt.hexdigest()
```
chamini2, line 99: should we run this only once in the `__post_init__` and store the result? to mimic the past `cached_property`?

chamini2, line 99: should we run this only once in the `__post_init__` and store the result? to mimic the past `cached_property`?

isidentical: I've reduced the number of `_key` accesses to one so it is probably not necessary.

src/fal/packages/_run_hook.py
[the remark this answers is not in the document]
isidentical: I've unified both fal.cli and this part, so once this is landed and the changes are applied it will fix both of them I think. Thanks for mentioning about this btw!

src/fal/packages/environment.py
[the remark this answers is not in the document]
isidentical: The parsing of dependencies would be a bit tricky (`fal==5.0`, `fal!=5.0`, and a lot of other variations) though we can just let pip handle it. It already raises a warning if you have a dependency conflict, and I think it will handle this case too. If it starts creating confusions, we can think about it later I think.
 
```console
 $ pip install pyjokes==0.5.0 pyjokes==0.6.0                                                                                                                                                                                                                                         5h54m
Collecting pyjokes==0.5.0
  Using cached pyjokes-0.5.0-py2.py3-none-any.whl (12 kB)
ERROR: Cannot install pyjokes==0.5.0 and pyjokes==0.6.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested pyjokes==0.5.0
    The user requested pyjokes==0.6.0
```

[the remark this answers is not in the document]
chamini2: Cool, let's do that!

src/fal/packages/environment.py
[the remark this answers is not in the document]
isidentical: Done! (and used a different method for the rest [fal, dbt-core] so we can drop the dependency on importlib-metadata).

src/fal/packages/environment.py
isidentical: @chamini2 I just tried and noticed that since the dataclass if frozen, we can't do it on the `__post_init__` 😄 As I said, it is extremely fast and side-effect free, so let's not spend too much time on it.

isidentical:
Thanks a lot for the reviews guys! @chamini2 please re-review, all the reviews should be addressed now.

src/fal/packages/environment.py
```diff
@@ -0,0 +1,201 @@
...
+            yield plugin_name
+
+
+def _get_default_requirements() -> Iterator[Tuple[str, Optional[str]]]:
+    import pkg_resources
+    import dbt.version
+
+    # TODO: for dev versions, we need a better solution.
+    if _IS_DEV_VERSION:
+        yield "git+https://github.com/fal-ai/fal@batuhan/fea-318-use-the-environment-option-for-local", None
```
chamini2, line 82: Probably needs to change before merge?

src/fal/packages/environment.py
```diff
@@ -0,0 +1,201 @@
...
+_BASE_CACHE_DIR = Path(user_cache_dir("fal", "fal"))
+_BASE_CACHE_DIR.mkdir(exist_ok=True)
+
+_BASE_VENV_DIR = _BASE_CACHE_DIR / "venvs"
+_BASE_VENV_DIR.mkdir(exist_ok=True)
+
+# TODO: for dev versions, we need a better solution.
+_IS_DEV_VERSION = True
+
+# This is from DBT itself
```
chamini2, line 46: Add a link to the file in the repo
```suggestion
# NOTE: This is from dbt https://github.com/dbt-labs/dbt-core/blob/7bd861a3514f70b64d5a6c642b4204b50d0d3f7e/core/dbt/version.py#L209-L236
```

chamini2 (approved):
Awesome, I think that is it!

src/fal/packages/environment.py
isidentical: Yep, awaiting the proper versions PR

isidentical: Fixed!

PR #490, opened 2022-07-29
fix: filter source nodes from nodes to run

### LINKED ISSUE #486: Stop trying to run sources in dbt runner
**Describe the bug**
fal takes sources as valid nodes to run in dbt and tries to run them.
It mostly ends up in a no-op operation, but if the source shares name with a model it can run the wrong model.

**Expected behavior**
fal should not try to run sources in dbt.

**Screenshots**

image

chamini2:
Fixes #486


isidentical (approved):
Logic LGTM. Looks like the test suite needs some changes to update the `NodeGraph`.

PR #492, opened 2022-08-01
feat: model-scoped environments

isidentical:
Requires #489, this PR implements support for isolation in the scope of individual models. Each model can now define an `environment:` setting directly under `fal:` block, and that setting will be applied to both Python model itself (if there is any) and all the hooks below it (unless they manually override).

### DIFF
--- a/src/fal/packages/_run_hook.py
+++ b/src/fal/packages/_run_hook.py
@@ -13,10 +13,11 @@ def run_fal_hook(
     fal_dbt_config: Dict[str, Any],
     arguments: Dict[str, Any],
     run_index: int,
+    hook_type: str,
     disable_logging: bool,
 ) -> None:
     from fal.node_graph import DbtModelNode, NodeGraph
-    from fal.planner.tasks import FalLocalHookTask
+    from fal.planner.tasks import FalLocalHookTask, HookType
     from faldbt.project import FalDbt
 
     fal_dbt = FalDbt(**fal_dbt_config)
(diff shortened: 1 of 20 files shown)


src/fal/packages/_run_hook.py
```diff
@@ -13,18 +13,28 @@ def run_fal_hook(
...
 
     fal_dbt = FalDbt(**fal_dbt_config)
     node_graph = NodeGraph.from_fal_dbt(fal_dbt)
     flow_node = node_graph.get_node(bound_model_name)
     assert isinstance(flow_node, DbtModelNode)
 
-    task = FalLocalHookTask(path, flow_node.model, arguments)
+    hook_type = HookType(hook_type)
+    if hook_type is HookType.MODEL_SCRIPT:
+        flow_node.model.python_model = Path(path)
```
chamini2, line 30: Shouldn't this be already encoded in the node_graph?

src/fal/planner/executor.py
```diff
@@ -27,7 +34,10 @@ def _collect_nodes(groups: List[TaskGroup], fal_dbt: FalDbt) -> Iterator[str]:
     for group in groups:
         if isinstance(group.task, DBTTask):
             yield from group.task.model_ids
-        if isinstance(group.task, FalLocalHookTask) and not group.task.is_hook:
+        if (
+            isinstance(group.task, FalLocalHookTask)
+            and not group.task.hook_type is HookType.HOOK
```
chamini2, line 39: From the changes in `_run_hook.py`, couldn't it happen that:

```py
isinstance(group.task, FalLocalHookTask) is True
and group.task.hook_type is HookType.MODEL_SCRIPT
```

Making this condition check `True` and we are assuming they are before/after scripts?

chamini2 (commented):
I have a couple of questions, looks very good in general

src/fal/planner/executor.py
isidentical: Good catch, it makes sense to filters this to only include scripts.

src/fal/planner/tasks.py
```diff
@@ -40,6 +39,12 @@ def execute(self, args: argparse.Namespace, fal_dbt: FalDbt) -> int:
         raise NotImplementedError
 
 
+class HookType(Enum):
+    HOOK = "HOOK"
+    SCRIPT = "SCRIPT"
```
chamini2, line 44: this is before or after script, right?

chamini2 (approved):
LGTM

src/fal/planner/tasks.py
isidentical: Yep!

PR #499, opened 2022-08-04
feat: enable fal runtime

mederka:


src/fal/planner/tasks.py
```diff
@@ -89,9 +89,16 @@ def _run_script(script: FalScript) -> Dict[str, Any]:
     # DBT seems to be dealing with only UTC times
     # so we'll follow that convention.
     started_at = datetime.now(tz=timezone.utc)
+
     try:
-        with _modify_path(script.faldbt):
+        from fal_runtime.client import runtime_run
+    except:
+        def runtime_run(script: FalScript):
```
chamini2, line 96: Not super sure a user always wants to use the runtime if the package is installed. Should we check a fal_project.yml property? a `~/.dbt/profiles.yml`? Somwhere?

isidentical:
ATL: How does fal runtime work interact with the last changes in the hook-system? Can it be another hook, like this
https://github.com/fal-ai/fal/blob/d3c5ab9803a7bdde6a31f55de95f8cb5babc71c9/src/fal/fal_script.py#L29-L33

and we could run it in the fal-runtime like how we run it in a separate environment (a separate hook task):
https://github.com/fal-ai/fal/blob/d3c5ab9803a7bdde6a31f55de95f8cb5babc71c9/src/fal/planner/tasks.py#L228-L245

PR #500, opened 2022-08-04
fix: ensure broken DBT models are not corrupting other models

isidentical:


integration_tests/projects/014_broken_dbt_models/dbt_project.yml
```diff
@@ -0,0 +1,13 @@
...
+profile: "fal_test"
+source-paths: ["models"]
+data-paths: ["data"]
+snapshot-paths: ["snapshots"]
+target-path: "{{ env_var('temp_dir') }}/target"
+vars:
+  fal-scripts-path: "scripts"
+
+models:
+  +schema: custom
```
chamini2, line 13: ATL: why are we setting this?

integration_tests/features/python_nodes.feature
```diff
@@ -45,3 +45,14 @@ Feature: Python nodes
...
+  Scenario: Broken DBT model leads to corruption of the dependant Python model
+    Given the project 014_broken_dbt_models
+    When the following command is invoked:
+      """
+      fal flow run --profiles-dir $profilesDir --project-dir $baseDir --select get_data regular_model
+      """
+    Then the following models are calculated:
+      | regular_model | get_data.py |
+    And the following scripts are ran:
+      | regular_model.post_hook.py |
```
chamini2, line 58: You could add a "and the exception X is thrown" which right now is a no-op. But we will make work eventually.

chamini2 (approved):

integration_tests/projects/014_broken_dbt_models/dbt_project.yml
isidentical: No specific reason, this project was copied directly from a different one so we should probably clear all of them.

integration_tests/features/python_nodes.feature
isidentical: This shouldn't supposed to a throw an exception.

integration_tests/projects/014_broken_dbt_models/dbt_project.yml
chamini2: I think this is tested in another project, but it's OK.

integration_tests/features/python_nodes.feature
chamini2: I thought we were running a failing run? But now I see we are doing `--select get_data regular_model`. Why aren't we running `process_data`? What is this actually testing?

isidentical: This test is ensuring that when there is a problematic model down in the chain (it doesn't really matter if it is selected or not, since this happens way before we exclude stuff in the graph) it doesn't affect the execution of others (just like how it doesn't effect it for DBT itself).

chamini2: But then it is not about the model being broken. It's just about the model not overwriting the dependencies as dbt models when they are actually fal models, right?

isidentical: > But then it is not about the model being broken. It's just about the model not overwriting the dependencies as dbt models when they are actually fal models, right?

Weirdly no. Somehow the model's broken status down in the chain affects the model listing / dependency order.

PR #501, opened 2022-08-04
feat: pass through full-refresh flag to dbt calls

chamini2:


integration_tests/projects/008_pure_python_models/models/staging/model_c.py
```diff
@@ -1,29 +1,15 @@
...
-output = f"Model name: {model_name}"
-output = output + f"\nStatus: {context.current_model.status}"
-
-output = output + "\nModel dataframe name: {model_name}"
-temp_dir = os.environ["temp_dir"]
-
-write_dir = open(reduce(os.path.join, [temp_dir, model_name + ".txt"]), "w")
-write_dir.write(output)
-write_dir.close()
+create_dynamic_artifact(context)
```
isidentical, line 15: AFAIK `create_dynamic_artifact` only creates artifacts for scripts (e.g., `model.script_name.txt` instead of `model.txt`), is this intended?

chamini2: Thanks for pointing it out!

isidentical (approved):
🚀 LGTM

PR #505, opened 2022-08-08
feat: implement "local" hooks

isidentical:
Implements the ability to specify the local (user's own environment) for hooks.

### DIFF
--- a/src/faldbt/parse.py
+++ b/src/faldbt/parse.py
@@ -17,7 +17,7 @@ from dbt.logger import GLOBAL_LOGGER as logger
 from faldbt.utils.yaml_helper import load_yaml
 
 if TYPE_CHECKING:
-    from fal.packages.environment import Environment
+    from fal.packages.environment import BaseEnvironment
 
 FAL_SCRIPTS_PATH = "fal-scripts-path"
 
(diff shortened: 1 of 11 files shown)


src/faldbt/parse.py
```diff
@@ -174,22 +174,27 @@ def get_global_script_configs(source_dirs: List[Path]) -> Dict[str, List[str]]:
...
-    else:
-        raise FalParseError(
-            f"Could not find the environment definition for {name} in {fal_project_path}"
-        )
+        if "name" not in environment:
+            raise FalParseError("Environment must have a name")
+
+        name = environment.get("name")
+        if _is_local_environment(name):
+            raise FalParseError(f"Environment name conflicts with a reserved name: {name}.")
```
chamini2, line 194: How do we feel of making the `local` environment actually a `null`?

https://www.educative.io/answers/how-to-represent-null-values-in-yaml

```yaml
        post-hook:
          - path: environment_type_2.py
            environment: null
```


```yaml
        post-hook:
          - path: environment_type_2.py
            environment: ~
```


```yaml
        post-hook:
          - path: environment_type_2.py
            environment:
```

chamini2: I do like how `local` is explicit. But I don't know, just a thought.

chamini2 (approved):
preeetty cool

src/faldbt/parse.py
isidentical: @mederka @turbo1912 @burkaygur any preferences? It is kind of annoying that this prevents people from naming their own environments `local` but I feel just like @chamini2 in the sense that it is more explicit than `~`.

isidentical: Let's merge this as is, and if anyone has any extra thoughts on this we can address them separately on https://linear.app/features-and-labels/issue/FEA-372/decide-on-local-vs-for-local-environments.

PR #507, opened 2022-08-08
ci: fix tag sent to git-cliff

chamini2:


turbo1912 (approved):

.github/workflows/python-release.yml
```diff
@@ -27,19 +38,30 @@ jobs:
...
+
           poetry build
 
       - name: Generate a changelog
         uses: orhun/git-cliff-action@v1
         id: git-cliff
         with:
           config: cliff.toml
-          args: v${{ env.prev_version_tag }}..HEAD
+          args: ${{ env.prev_version_tag }}..HEAD
```
isidentical, line 64: Ah, nice catch!

isidentical (approved):
🤖  thanks for the fix (and nice idea on the unification)!

PR #509, opened 2022-08-09
feat: support singular tests

### LINKED ISSUE #371: Add support for singular tests
**Context**
dbt has two types of tests: singular and generic. Fal currently supports only generic tests. We need support for singular tests as well.

 A list of singular tests should be accessible to the user, either via a dedicated `list_singular_tests` function or `context.singular_tests`.

### LINKED ISSUE #497: Better support for singular tests
**Describe the bug**
Fal object model does not seem to support singular tests. Getting various errors with dbt project that uses custom singular tests (standalone sql tests).

```py
from fal import FalDbt
faldbt = FalDbt(profiles_dir='.dbt', project_dir='/workspace')
print(faldbt.tests[0])
# Getting AttributeError: 'DbtTest' object has no attribute 'column'
for t in faldbt.tests:
    print(t.name)
# Geeting a AttributeError: 'ParsedSingularTestNode' object has no attribute 'test_metadata'
```

**Your environment**
```bash
# fal --version
fal 0.4.1
# dbt --version
Core:
  - installed: 1.2.0
  - latest:    1.2.0 - Up to date!
Plugins:
  - snowflake: 1.2.0 - Up to date!
```
**How to reproduce**
We have a dbt project with generic and custom tests, running `dbt run` and then code above

**Expected behavior**
Expect to be able to retrieve items from tests collection and be able to work with them regardless of type of test

**Actual behavior**
Can not work with tests in fal

chamini2:
# TODO
- [x] tests

closes #371 
closes #497

### DIFF
--- a/integration_tests/features/environment.py
+++ b/integration_tests/features/environment.py
@@ -20,12 +20,10 @@ def after_scenario(context, scenario):
 
     if hasattr(context, "exc") and context.exc:
         from traceback import print_exception
-        from sys import _OptExcInfo
 
-        exc_info: _OptExcInfo = context.exc
-        _etype, exception, _tb = exc_info
+        _etype, exception, _tb = context.exc
 
-        print_exception(*exc_info)
+        print_exception(*context.exc)
 
         raise AssertionError("Should have expected exception") from exception
 
--- a/integration_tests/features/environment.py
+++ b/integration_tests/features/environment.py
@@ -20,10 +20,14 @@ def after_scenario(context, scenario):
 
     if hasattr(context, "exc") and context.exc:
         from traceback import print_exception
+        from sys import _OptExcInfo
 
-        print_exception(*context.exc)
-        _etype, exception, _tb = context.exc
-        raise exception
+        exc_info: _OptExcInfo = context.exc
+        _etype, exception, _tb = exc_info
+
+        print_exception(*exc_info)
+
+        raise AssertionError("Should have expected exception") from exception
 
 
 def before_tag(context, tag):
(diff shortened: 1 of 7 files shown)


integration_tests/features/environment.py
```diff
@@ -20,10 +20,14 @@ def after_scenario(context, scenario):
 
     if hasattr(context, "exc") and context.exc:
         from traceback import print_exception
+        from sys import _OptExcInfo
```
isidentical, line 23: never seen this before

isidentical: Is it coming from runtime?
```py
>>> from sys import _OptExcInfo
  File "<stdin>", line 1, in <module>
ImportError: cannot import name '_OptExcInfo' from 'sys' (unknown location)
```

src/faldbt/project.py
```diff
@@ -86,45 +91,81 @@ def _set_status(self, status: str):
...
+            return self.model_ids[0]
+
+    # TODO: Deprecate?
+    @property
+    def source(self):
+        if self.source_id:
+            parts = self.source_id.split(".")
+            return parts[-2], parts[-1]
+
+    # TODO: Deprecate?
```
isidentical, line 145: I'd say yes, the model_id/source_id APIs seems enough. All the post processing can go to client code. Though no strong opinions, it is very light so it can stay either.

isidentical (approved):

integration_tests/features/environment.py
chamini2: It's the type of a [`sys.exc_info()`](https://docs.python.org/3/library/sys.html#sys.exc_info) call. I can remove it just to be sure, though.

src/faldbt/project.py
chamini2: We can solve it later, it is lightweight.

isidentical: Yep!

PR #513, opened 2022-08-10
feat: add support for sources meta property access in scripts context

### LINKED ISSUE #506: Get metadata on sources
**Context**
To use fal-ai to notify users via slack on failed source tests we would like to use the owner configured within the `metadata` tag, the same way as it's possible for models

**Is your feature request related to a problem? Please describe.**
No

To have a metadata parameter for each source (callable via list_sources() ) the same way it's already possible with models

The only possible, but not viable, alternative I could think of, would be to hardcode it into the script based on the model name

**Additional context**

chamini2:
closes #506

### DIFF
--- a/integration_tests/features/meta_object.feature
+++ b/integration_tests/features/meta_object.feature
@@ -1,4 +1,4 @@
-Feature: `write_to_source` function
+Feature: Access to meta object from schema
   Background: Project Setup
     Given the project 005_functions_and_variables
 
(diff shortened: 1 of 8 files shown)


integration_tests/features/meta_object.feature
```diff
@@ -0,0 +1,28 @@
+Feature: `write_to_source` function
```
isidentical, line 1:
```suggestion
Feature: Accessing to source-level meta properties
```

isidentical (approved):
Can't see anything wrong, well done!

integration_tests/features/meta_object.feature
chamini2: good catch!

PR #517, opened 2022-08-22
refactor: use sqlalchemy engine to write and read dataframes by default

chamini2:
adapters to test:
- bigquery
- snowflake
- postgres

### DIFF
--- a/src/faldbt/lib.py
+++ b/src/faldbt/lib.py
@@ -251,6 +251,9 @@ def fetch_target(
 
 
 def _fetch_relation(adapter: SQLAdapter, relation: BaseRelation) -> pd.DataFrame:
+    if adapter.type() == "postgres":
+        return _sqlalchemy_engine_fetch_relation(adapter, relation)
+
     query = f"SELECT * FROM {relation}"
     return _execute_sql(adapter, query)[1]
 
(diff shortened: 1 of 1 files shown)


src/faldbt/lib.py
```diff
@@ -499,6 +502,18 @@ def null_dump(sql, *multiparams, **params):
     return sqlalchemy.create_mock_engine(url_string, executor=null_dump)
 
 
+def _create_engine_from_connection(adapter: SQLAdapter):
+    if adapter.type() == "postgres":
+        url_string = "postgresql+psycopg2://"
+    else:
+        # TODO: add special cases as needed
+        logger.warn("No explicit url string for adapter {}", adapter.type())
+        url_string = f"{adapter.type()}://"
```
isidentical, line 511: Can we use `adapter.credentials.connection_string` here (since it would also contain the prefix)?

isidentical (approved):
Looks great, thanks! Quick question below

mederka (approved):

src/faldbt/lib.py
chamini2: `adapter.credentials.connection_string` is this a thing?!

isidentical: I recall seeing it somewhere, but now looking deeply I can't find it either. Probably misremembered :/

PR #519, opened 2022-08-25
feat: two-way bridge for transporting objects to isolated hooks

isidentical:


isidentical:
This is how I use the debugging interface

[asciicast](https://asciinema.org/a/d2yAFhStKo0EUCaOmz3cVTZWr)

src/fal/packages/bridge.py
```diff
@@ -0,0 +1,50 @@
...
+from functools import partial
+from multiprocessing.connection import Client, ConnectionWrapper, Listener
+from typing import Union
+
+import dill
+
+dill_wrapper = partial(ConnectionWrapper, dumps=dill.dumps, loads=dill.loads)
+
+
+class _DillListener(Listener):
```
mederka, line 29: I think it would be great to do a presentation on how this listener works.

turbo1912: +1

src/fal/utils.py
```diff
@@ -25,3 +26,25 @@ def next(self) -> int:
...
+        super().__init__(*args, **kwargs)
+
+    def __getstate__(self):
+        return self._serialization_state
+
+    def __setstate__(self, state):
+        super().__init__(*state["args"], **state["kwargs"])
+
+
+def has_side_effects(cls):
```
turbo1912, line 46: if we are just using this internally and the only purpose of this is to decide on a serilization strategy, should the name also reflect that? `reinitilize` might be a better name maybe?

src/fal/packages/isolated_runner.py
```diff
@@ -0,0 +1,92 @@
...
+        logger.debug("Received the callable at {}", address)
+        try:
+            result = callable()
+            exception = None
+        except BaseException as exc:
+            result = None
+            exception = exc
+        finally:
+            connection.send((result, exception))
+        return result
```
turbo1912, line 63: let's do a demo of this, not sure if I understand how the breakpoint resumes

turbo1912 (approved):
did some drive by comments :D

mederka (approved):
Great job, @isidentical!

PR #522, opened 2022-09-03
fix: correctly serialize script type when constructing task

isidentical:
Seems like this went unnoticed because we can't create empty models locally when we were trying to run the test `features/run.feature:128`. Should be fixed now.


mederka (approved):

chamini2 (approved):
Are the 2 changes related or 2 unrelated fixes?

isidentical:
Related. Fixing the test makes it fail on all backends (not just bigquery) and the fix in the serialization logic fixes the CI (and the actual bug).

PR #523, opened 2022-09-05
refactor: define the explicit flow between env<->conn

isidentical:
This is a pre-requisite for the conda implementation, pretty much a decoupling of the IPC protocol from the virtual env system (as well as other useful refactors)

### DIFF
--- a/src/fal/packages/bridge.py
+++ b/src/fal/packages/bridge.py
@@ -32,6 +32,9 @@ class _DillListener(Listener):
 
 
 def controller_connection() -> _DillListener:
+    # The controller here assumes that there will be at most one
+    # client. This restriction might change in the future as an
+    # optimization.
     return _DillListener()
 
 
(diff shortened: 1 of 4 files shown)


src/fal/packages/bridge.py
```diff
@@ -32,6 +32,9 @@ def accept(self) -> ConnectionWrapper:
 
 
 def controller_connection() -> _DillListener:
+    # The controller here assumes that there will be at most one
+    # client. This restriction might change in the future as an
+    # optimization.
```
chamini2, line 37: Should we do a singleton? I mean to actually make sure it there is only one.

chamini2 (approved):
To be honest, I think I have to look deeper into some of the implementation details of other PRs, but looks great for now!

mederka (approved):
Looks great!

src/fal/packages/bridge.py
isidentical: This is not a singleton though, there might be multiple controller connections but each one of them will only wait for one child to connect (N controllers, N children. Instead of 1 controller and N children in the future if we ever come to that point).

PR #524, opened 2022-09-07
refactor: move to new logging system

chamini2:


src/fal/logger.py
```diff
@@ -0,0 +1,53 @@
+from dataclasses import dataclass
+
+from dbt.logger import log_manager
+from dbt.events.functions import fire_event
```
mederka, line 4: Can we rename this function when we use it? I think calling `fire_event` each time we want to log is a bit awkward.

mederka (commented):
Great job, @chamini2 !

I have one question. Instead of calling fire_event and passing a message object there, why don't build something like a `fal_logger` and and add methods to it, like `warn`, `debug` etc? Maybe I'm missing something.

chamini2:
What do you think now @mederka ?

src/fal/logger.py
```diff
@@ -0,0 +1,119 @@
...
+        else:
+            return msg
+
+
+# TODO: do we still need to do this?
+def reconfigure_logging() -> None:
+    # Disabling the dbt.logger.DelayedFileHandler manually
+    # since we do not use the new dbt logging system
+    # This fixes issue https://github.com/fal-ai/fal/issues/97
+    log_manager.set_path(None)
```
mederka, line 64: What happens if we don't run this?

mederka (approved):
👏 👏 👏

src/fal/logger.py
```diff
@@ -0,0 +1,119 @@
...
+from dbt.events.base_types import (
+    TestLevel as _TestLevel,
+    DebugLevel as _DebugLevel,
+    InfoLevel as _InfoLevel,
+    WarnLevel as _WarnLevel,
+    ErrorLevel as _ErrorLevel,
+)
+
+
+class LOGGER:
```
isidentical, line 14: ATL: why is the logger is all capitalized?

src/fal/logger.py
```diff
@@ -0,0 +1,119 @@
...
+    @classmethod
+    def test(cls, msg: str, *args, **kwargs):
+        fire_event(TestMessage(_prepare_msg(msg, *args, **kwargs)))
+
+    @classmethod
```
isidentical, line 19: WDYT about making `Logger` a normal class and having a single instance like `logger = Loggger()` at the end of the file (or `LOGGER`)? We could even have different `Logger` classes and having it configurable seems like a win (e.g. `logger` would now point to the `IPCLogger()`)., e.g. a logger class that uses a `fire_event_to_the_bridge` method etc.

src/fal/logger.py
chamini2: because we never instantiate it. Maybe this is not the correct pattern to follow and we should instead offer the functions directly?

Then importing and using like:

```
import fal.logger as logger

logger.info(...)
```

src/fal/logger.py
chamini2: This is a valid approach too. We just don't really hold anything in the class.

src/fal/logger.py
```diff
@@ -0,0 +1,119 @@
...
+    if args:
+        if kwargs:
+            return msg.format(*args, **kwargs)
+        else:
+            return msg.format(*args)
+    else:
+        if kwargs:
+            return msg.format(**kwargs)
+        else:
+            return msg
```
isidentical, line 56: ATL: Does this work or is there a scenario where the logic above is needed?
```suggestion
def _prepare_msg(msg: str, *args, **kwargs):
    return msg.format(*args, **kwargs)
 ```

File: src/fal/packages/environments/base.py [hunk 6]
```diff
@@ -40,7 +39,20 @@ def rmdir_on_fail(path: Path) -> Iterator[None]:
...
+    elif kind.startswith("debug"):
+        LOGGER.debug(message, *args, **kwargs)
+    elif kind.startswith("info"):
+        LOGGER.info(message, *args, **kwargs)
+    elif kind.startswith("warn"):
+        LOGGER.warn(message, *args, **kwargs)
+    elif kind.startswith("error"):
+        LOGGER.error(message, *args, **kwargs)
+    else:
+        LOGGER.info(message, *args, **kwargs)
```
isidentical, line 55:
```suggestion
    log_method = getattr(LOGGER, kind)
    log_method(message, *args, **kwargs)
```

src/fal/logger.py
isidentical: Both seems to work for me, but if we are using it just for the namespace-ing I'd be also fine with a single instance.

src/fal/logger.py
chamini2: There is a scenario. When we log the ipynb files we sometimes get `strings with {braces}` and then the `format` call expects to fill them with something. So we have to test if something WAS passed. I just thought of a small improvement, though.

src/fal/packages/environments/base.py
chamini2: We lose the `startswith` check (think `warning` vs `warn`). Also, we would FAIL instead of defaulting to `info`. Maybe failing is better so the developer fixes the error...

isidentical:
Amazing work @chamini2 🚀

PR #525, opened 2022-09-08
ci: namespace integration tests to run in parallel with same datawarehouse

chamini2:
Test bigquery and snowflake


chamini2:
Enabled  `Default table expiration     1 day` for BigQuery. We have to look into Snowflake and others we might add.

Maybe add a post-processing step that just deletes the whole thing?

In that case, I would go for making the namespacing happen in schema level, not table name. I did not go for that right now because I think some adapters fail if the schema does not already exist.

turbo1912 (approved):

chamini2:
Snowflake has [transient tables](https://docs.snowflake.com/en/user-guide/tables-temp-transient.html#transient-tables) which are not the same, but better than standard ones.

I can also just look into dropping all of the tables at the end.

isidentical (approved):

PR #527, opened 2022-09-09
fix: ensure the venv based environments are thread safe

isidentical:


isidentical:
Seems like an oversight from the previous refactoring PR.

PR #528, opened 2022-09-09
feat: implement initial support for conda

isidentical:
Closes FEA-400

### DIFF
--- a/integration_tests/projects/013_structured_hooks/fal_project.yml
+++ b/integration_tests/projects/013_structured_hooks/fal_project.yml
@@ -1,11 +1,23 @@
 environments:
   - name: not-funny
+    type: venv
+
   - name: funny
+    type: venv
     requirements:
       - pyjokes==0.6.0
+
   - name: pyjokes-0.5.0
+    type: venv
     requirements:
       - pyjokes==0.5.0
+
   - name: pyjokes-0.6.0
+    type: venv
     requirements:
       - pyjokes==0.6.0
+
+  - name: funny-conda
+    type: conda
+    packages:
+      - pyjokes=0.6.0
--- a/src/fal/packages/dependency_analysis.py
+++ b/src/fal/packages/dependency_analysis.py
@@ -1,7 +1,11 @@
 from __future__ import annotations
 
 from pathlib import Path
-from typing import Iterator, Optional, Tuple
+from typing import Iterator, List, Optional, Tuple
+
+import importlib_metadata
+
+from fal.utils import cache_static
 
 import importlib_metadata
 
@@ -81,3 +85,11 @@ def _get_fal_package_name() -> Tuple[str, Optional[str]]:
 def get_default_requirements() -> Iterator[Tuple[str, Optional[str]]]:
     yield _get_fal_package_name()
     yield from _get_dbt_packages()
+
+
+@cache_static
+def get_default_pip_dependencies() -> List[str]:
+    return [
+        f"{package}=={version}" if version else package
+        for package, version in get_default_requirements()
+    ]
--- a/src/fal/packages/environments/__init__.py
+++ b/src/fal/packages/environments/__init__.py
@@ -1,8 +1,22 @@
-from typing import List
+from pathlib import Path
+from typing import List, Optional, Dict, Any
 
 from fal.packages.environments.base import BASE_CACHE_DIR, BaseEnvironment
+from fal.packages.environments.conda import CondaEnvironment
 from fal.packages.environments.virtual_env import VirtualPythonEnvironment
 
+REGISTERED_ENVIRONMENTS: Dict[str, BaseEnvironment] = {
+    "conda": CondaEnvironment,
+    "venv": VirtualPythonEnvironment,
+}
 
-def create_environment(*, requirements: List[str]) -> BaseEnvironment:
-    return VirtualPythonEnvironment(requirements=requirements)
+
+def create_environment(name: str, kind: str, config: Dict[str, Any]) -> BaseEnvironment:
+    env_type = REGISTERED_ENVIRONMENTS.get(kind)
+    if env_type is None:
+        raise ValueError(
+            f"Invalid environment type (of {kind}) for {name}. Please choose from: "
+            + ", ".join(REGISTERED_ENVIRONMENTS.keys())
+        )
+
+    return env_type.from_config(config)
--- a/src/fal/packages/environments/conda.py
+++ b/src/fal/packages/environments/conda.py
@@ -1,0 +1,141 @@
+from __future__ import annotations
+
+import hashlib
+import os
+import shutil
+import subprocess
+import sysconfig
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import ContextManager, List, Dict, Any
+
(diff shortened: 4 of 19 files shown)


src/fal/packages/dependency_analysis.py
```diff
@@ -60,3 +61,11 @@ def get_default_requirements() -> Iterator[Tuple[str, Optional[str]]]:
 
     for adapter_name, adapter_version in _get_dbt_plugins_info():
         yield f"dbt-{adapter_name}", adapter_version
+
+
+@cache_static
```
mederka, line 66: ATL: what does this do?

src/fal/packages/environments/__init__.py
```diff
@@ -1,8 +1,19 @@
...
+def create_environment(*, pip_requirements: Optional[List[str]] = None, conda_requirements: Optional[List[str]] = None) -> BaseEnvironment:
+    if conda_requirements is not None:
+        if pip_requirements:
+            raise RuntimeError("Pip based installations are not supported when using conda environments.")
+        return CondaEnvironment(conda_requirements)
+    elif pip_requirements is not None:
+        assert not conda_requirements
+        return VirtualPythonEnvironment(pip_requirements)
+    else:
+        raise RuntimeError("Either provide requirements (for pip) or conda_requirements.")
```
mederka, line 19: What do we do if a script/model has no package requirements?

src/fal/packages/dependency_analysis.py
isidentical: Think of it as a lazy global, until the function is called we don't construct anything and once there is at least one call we always return the same value (so pretty-much singleton-izing the thing).

src/fal/packages/environments/__init__.py
isidentical: This happens when you define an environment and that environment doesn't have any dependencies, like the following:
```yaml
environments:
  - name: not-funny
```

if you don't want to include any dependencies, you can still do it by using an empty list literal (YAML syntax)
```yaml
    # This is going to construct an empty (yet still isolated) environment.
    requirements: []
```

integration_tests/projects/013_structured_hooks/fal_project.yml
```diff
@@ -9,3 +11,6 @@ environments:
   - name: pyjokes-0.6.0
     requirements:
       - pyjokes==0.6.0
+  - name: funny-conda
+    conda_requirements:
```
isidentical, line 15: What do you think of this name? Should we instead call it conda_deps? Maybe something else? (also should we now start calling `requirements` field, `pip_requirements`/`pip_deps` for the sake of consistency)

CC: @burkaygur @turbo1912 on this as well.

chamini2: > What do you think of this name? Should we instead call it conda_deps?

Follow what conda normally calls it. It seems to be packages, so we can do `conda_packages`.

> also should we now start calling requirements field, pip_requirements/pip_deps for the sake of consistency

I would agree with this. And there is no problem since this feature is not being used yet.

chamini2: Also to consider: what if we do like profiles.yml, you specify a type of env and the requirments/packages/xxx in a different property.

```yaml
  - name: funny-pip
    type: pip
    requirements:
      - pyjokes==0.6.0
  - name: funny-conda
    type: conda
    packages:
      - pyjokes==0.6.0
```

I think this may be good for a future where env types have different properties to set (ram, gpu, etc)

isidentical: I think this is even better, since it would also reduce the ambiguity on defaults (e.g. you would be able to just specify `type: fal-cloud` which would have a default machine type, instead of specifying `machine` explicitly as we are doing now).

src/fal/packages/environments/virtual_env.py
```diff
@@ -2,34 +2,35 @@
...
+from fal.utils import cache_static
 
 _BASE_VENV_DIR = BASE_CACHE_DIR / "venvs"
 _BASE_VENV_DIR.mkdir(exist_ok=True)
 
 
 @dataclass
 class VirtualPythonEnvironment(BaseEnvironment[Path], make_thread_safe=True):
-    requirements: List[str] = field(default_factory=list)
+    pip_requirements: List[str] = field(default_factory=list)
```
chamini2, line 29: ATL: would we change this?

src/fal/utils.py
```diff
@@ -48,3 +49,20 @@ def has_side_effects(cls):
     make the regular serialization problematic (e.g. registering
     adapters to DBT's global factory)."""
     return type(cls.__name__, (_ReInitialize, cls), {})
+
+
+_NO_HIT = object()
+
+
+def cache_static(func):
+    """Cache the result of an parameter-less function."""
```
chamini2, line 58: cool!

src/fal/packages/environments/conda.py
```diff
@@ -0,0 +1,125 @@
...
+def get_conda_executable() -> Path:
+    for path in [_FAL_CONDA_HOME, None]:
+        conda_path = shutil.which(_CONDA_COMMAND, path=path)
+        if conda_path is not None:
+            return conda_path
+    else:
+        # TODO: we should probably point to the instructions on how you
+        # can install conda here.
+        raise RuntimeError(
+            "Could not find conda executable. Please install conda or set FAL_CONDA_HOME."
```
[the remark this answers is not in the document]
isidentical: Currently it might, but we don't explicitly support mamba (it is on the todo list though). It might require a bit of more extra work before we have complete support but yes, in theory you can put any executable under `FAL_CONDA_HOME`. The primary use case is for people who don't have a global conda installation (like me) since that tends to shadow everything else.

src/fal/utils.py
isidentical: Thx!

src/fal/packages/environments/virtual_env.py
isidentical: Fixed now, so venv uses `requirements` and conda uses `packages` with both in config and in the internal implementation.

src/fal/packages/environments/conda.py
```diff
@@ -0,0 +1,125 @@
...
+
+# Specify the path where the conda binary might
+# reside in (or mamba, if it is the preferred one).
+_CONDA_COMMAND = os.environ.get("CONDA_EXE", "conda")
+_FAL_CONDA_HOME = os.getenv("FAL_CONDA_HOME")
+
+
+@dataclass
+class CondaEnvironment(BaseEnvironment[Path], make_thread_safe=True):
+    conda_requirements: List[str]
```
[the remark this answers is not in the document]
isidentical: went with your `packages` suggestion.

integration_tests/projects/013_structured_hooks/fal_project.yml
isidentical: @chamini2 should be implemented now! I think it looks fantastic with your suggestion

chamini2 (approved):
⭐ ⭐ ⭐ ⭐ ⭐

mederka (approved):
This is awesome!

PR #529, opened 2022-09-12
fix: correct dependency analysis for external dbt packages

isidentical:


src/fal/packages/dependency_analysis.py
```diff
@@ -1,45 +1,26 @@
...
+def _get_dbt_packages() -> Iterator[Tuple[str, str]]:
+    # package_distributions will return a mapping of top-level package names to a list of distribution names (
+    # the PyPI names instead of the import names). An example distirbution info is the following, which
+    # contains both the main exporter of the top-level name (dbt-core) as well as all the packages that
+    # export anything to that namespace:
+    #   {"dbt": ["dbt-core", "dbt-postgres", "dbt-athena-adapter"]}
+    #
+    # This won't only include dbt.adapters.xxx, but anything that might export anything to the dbt namespace
+    # (e.g. a hypothetical plugin that only exports stuff to dbt.includes.xxx) which in theory would allow us
+    # to replicate the exact environment.
```
chamini2, line 17: Really interesting. I think this sounds fine, but I am curious if there will be complains or anything like that. Let's go with this and we will figure it out!

chamini2 (approved):
Thanks!

src/fal/packages/dependency_analysis.py
isidentical: We could also limit this to adapters only, but I don't see any reason for it right now. As you said, if in the future if we see this as a problem we can add that filter as well.

PR #534, opened 2022-09-14
feat: generalize dual Python IPC and speed-up conda builds

isidentical:
More description about the actual algoritm is under `DualPythonIPC` class.

### DIFF
--- a/src/fal/packages/environments/base.py
+++ b/src/fal/packages/environments/base.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
+import os
 import shutil
 import subprocess
+import sysconfig
 import threading
 from collections import defaultdict
 from contextlib import ExitStack, contextmanager, nullcontext
@@ -12,7 +14,7 @@ from typing import Any, Callable, ContextManager, Generic, Iterator, TypeVar, Di
 from platformdirs import user_cache_dir
 
 from fal.logger import LOGGER
-from fal.packages import bridge
+from fal.packages import bridge, isolated_runner
 
 BASE_CACHE_DIR = Path(user_cache_dir("fal", "fal"))
 BASE_CACHE_DIR.mkdir(exist_ok=True)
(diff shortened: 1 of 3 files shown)


src/fal/packages/environments/base.py
```diff
@@ -176,3 +178,71 @@ def poll_until_result(
...
+            env={"PYTHONPATH": python_path, **os.environ},
+        )
+
+
+def python_path_for(*search_paths) -> str:
+    assert len(search_paths) >= 1
+    return os.pathsep.join(
+        # sysconfig takes the virtual environment path and returns
+        # the directory where all the site packages are located.
+        sysconfig.get_path("purelib", vars={"base": search_path})
```
chamini2, line 234: ATL: what is purelib?

src/fal/packages/environments/conda.py
```diff
@@ -67,48 +65,18 @@ def _get_or_create(self) -> Path:
...
-                python_executable,
-                isolated_runner.__file__,
-                bridge.encode_service_address(service.address),
-            ],
-        )
+    def open_connection(self, conn_info: Path) -> DualPythonIPC:
+        primary_env = get_primary_virtual_env()
+        primary_env_path = primary_env.get_or_create()
+        secondary_env = conn_info
+        return DualPythonIPC(self, primary_env_path, secondary_env)
```
chamini2, line 79:
```suggestion
        secondary_env_path = conn_info
        return DualPythonIPC(self, primary_env_path, secondary_env_path)
```

chamini2 (approved):

src/fal/packages/environments/conda.py
isidentical: thx!

src/fal/packages/environments/base.py
isidentical: `purelib` points out to the path where the actual packages are located. E.g. my venv is under `~/.cache/fal/xxx` but the packages actually reside in `~/.cache/fal/xxx/lib/site-packages/something` (depending on the platform), using `purelib` with the base of our search_path (which is the venv directory, or conda) we can find the proper path on where all the site-packages are located.

isidentical: Definitions are encoded in here (and packaging tools like pip/conda use them) https://github.com/python/cpython/blob/c112489f877a37cd6e887f0893fc1f4843c168fa/Lib/sysconfig.py#L88-L118

PR #537, opened 2022-09-15
chore: add telemetric project identifier

mederka:


src/fal/cli/cli.py
```diff
@@ -16,6 +16,7 @@
 def cli(argv: List[str] = sys.argv):
     # Wrapper to be able to shutdown telemetry afterwards
     try:
+        telemetry.persist_hashed_project_name(argv[1:])
```
chamini2, line 19: I think this would not be called for using fal as a library, right?

```
from fal import FalDbt

faldbt = FalDbt(...)
faldbt.source(...) # no project set
```

Should we set this in `FalDbt.__init__` and use the `config` object which I think already has a hashing solution?

mederka: Good catch, @chamini2. My only concern would then be the telemetry call on line 26. What do you think about it?

chamini2: I wonder if we can _spare_ the project identifier for that call and we can "tie it" to the rest of the call by giving an `invocation_id` (do we have one of those?)

mederka: Ok, so, PR has been updated with limited scope as discussed.

chamini2 (approved):

PR #538, opened 2022-09-15
feat: add DO_NOT_TRACK as option for telemetry opt out

mederka:


src/fal/telemetry/telemetry.py
```diff
@@ -183,6 +183,10 @@ def check_stats_enabled():
     if "FAL_STATS_ENABLED" in os.environ:
         return os.environ["FAL_STATS_ENABLED"].lower() == "true"
 
+    if "DO_NOT_TRACK" in os.environ:
```
isidentical, line 186: Can you also add this to the docstring right above.

isidentical (approved):
Nice thinking on standardization.

PR #542, opened 2022-09-15
ci: only run generate typing job when originating from upstream

isidentical:
It seems like the PRs coming from my fork fail at this stage, it is not big since we don't receive much outside contributions but just to make the CI green I think we can ignore this job on forks.

### DIFF
--- a/.github/workflows/generate_typing.yml
+++ b/.github/workflows/generate_typing.yml
@@ -20,9 +20,10 @@ jobs:
       cancel-in-progress: true
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


isidentical:
From what I understood, it is for making the new commit trigger the CI: https://github.com/fal-ai/fal/pull/510#issue-1333816178

chamini2:
You are right! I just found an issue which seems to make this work: https://github.com/stefanzweifel/git-auto-commit-action/issues/211

I think we need to move to checkout@v3 and remove the `token` from it. Can we try that?

isidentical:
Trying it!

isidentical:
Seems to be working, thanks again @chamini2!

PR #549, opened 2022-09-23
misc: drop redundant dependencies to speed-up build times

isidentical:
### Integration tests

Adapter to test:
- postgres

Python version to test:

- 3.8

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,17 +21,20 @@ classifiers = [
 [tool.poetry.dependencies]
 python = ">=3.7.1, <3.11"
 pandas = "^1.3.4"
-arrow = "^1.2.0"
 pyarrow = "^6.0.1"
-pydantic = "^1.8.2"
 PyYAML = "^6.0"
 agate-sql = "^0.5.8"
 requests = "^2.27.1"
 posthog = "^1.4.5"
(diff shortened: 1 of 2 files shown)


mederka (approved):
ATL: how big is the speed-up?

isidentical:
> ATL: how big is the speed-up?

It is nothing major, just shaving a couple of seconds from the installation time. Still an easy/simple win.

pyproject.toml
```diff
@@ -21,17 +21,20 @@ classifiers = [
 [tool.poetry.dependencies]
 python = ">=3.7.1, <3.11"
 pandas = "^1.3.4"
-arrow = "^1.2.0"
```
chamini2, line 24: are we sure about this one?

chamini2:
Started this to make sure: https://github.com/fal-ai/fal/actions/runs/3113115123

pyproject.toml
isidentical: `arrow` on PyPI seems to be an unrelated library (to actual Apache Arrow) that does date processing, https://pypi.org/project/arrow/. I grepped the codebase, and seems like we don't use it at all. The PyArrow dependency is still here by the way (just right under that line)

PR #554, opened 2022-09-28
feat: added snowflake read and write

burkaygur:


adapter/src/dbt/adapters/fal/teleport/snowflake.py
```diff
@@ -0,0 +1,69 @@
...
+
+    assert adapter.type() == "snowflake"
+
+    _adapter: SnowflakeAdapter = adapter  # type: ignore
+
+    database: str = relation.database  # type: ignore
+    schema: str = relation.schema  # type: ignore
+    table: str = relation.identifier  # type: ignore
+
+    with new_connection(adapter, "fal-snowflake:write_df_to_relation") as conn:
```
burkaygur, line 47: This could be handled in the calling function `write_df_to_relation` in line 73 of adapter_support

chamini2:
I would not call the directory `teleport`, since `teleport` is about external storage. We can change it to `pandas` since this solution is focused on pandas, or some other directory name.

Same comment for #555

adapter/src/dbt/adapters/fal/teleport/snowflake.py
isidentical: Agreed. I think it would be a bit more clean if we centralize the connection handling to a higher scope.

chamini2: I think is ok here

PR #560, opened 2022-09-28
fix: Find conda binary even if it is directly set in FAL_CONDA_HOME

chamini2:
We passed some/path/conda and it did not work, when passed some/path it worked.

### Integration tests

Adapter to test:
- postgres

Python version to test:

- 3.8


mederka (approved):
⭐

PR #561, opened 2022-09-29
ci: run jaffle shop for act1

isidentical:
snowflake

### DIFF
--- a/.github/workflows/test_dbt_fal.yml
+++ b/.github/workflows/test_dbt_fal.yml
@@ -1,0 +1,167 @@
+name: Test dbt-fal adapter
+
+on:
+  pull_request:
+    types: [assigned, opened, synchronize, reopened]
+    paths-ignore:
+      - "examples/**"
+      - "docsite/**"
+      - "README.md"
+      - "LICENSE"
+      - "docker/**"
+
+  push:
+    branches: [act1]
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        profile: ["postgres", "bigquery", "snowflake", "duckdb"]
+        python: ["3.8"]
+        environment: ["local", "venv", "conda"]
+        # bigquery implementation fails due to concurrency, let's
+        # limit it to 1 for now.
+        exclude:
+          - profile: "bigquery"
+            environment: "venv"
+          - profile: "bigquery"
+            environment: "conda"
+
+    # Run only the latest commit pushed to PR
+    concurrency:
+      group: "${{ github.head_ref || github.run_id }}-${{ github.workflow }}-${{ matrix.profile }}-${{ matrix.python }}-${{ matrix.environment }}"
+      cancel-in-progress: true
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          path: 'fal'
+
+      - uses: actions/checkout@v2
+        with:
+          repository: fal-ai/dbt-core
+          ref: act1
+          path: 'dbt-core'
+
+      - uses: actions/checkout@v2
+        with:
+          repository: fal-ai/jaffle_shop_with_fal
+          ref: act1
+          path: 'jaffle_shop'
+
+      - name: Start Docker database
+        working-directory: fal/integration_tests
+        if: contains(fromJSON('["postgres"]'), matrix.profile)
+        run: docker-compose up -d
+
+      - name: Install conda
+        uses: s-weigand/setup-conda@v1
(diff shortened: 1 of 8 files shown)


.github/workflows/test_dbt_fal.yml
```diff
@@ -0,0 +1,167 @@
+name: Test dbt-fal adapter
+
+on:
+  pull_request:
+    types: [assigned, opened, synchronize, reopened]
```
chamini2, line 5:
```suggestion
    types: [assigned, opened, synchronize, reopened]
    branches: ['act1**']
```

.github/workflows/test_dbt_fal.yml
```diff
@@ -0,0 +1,167 @@
...
+      - uses: actions/checkout@v2
+        with:
+          repository: fal-ai/dbt-core
+          ref: act1
+          path: 'dbt-core'
+
+      - uses: actions/checkout@v2
+        with:
+          repository: isidentical/jaffle_shop_with_fal # change this to fal-ai/jaffle_shop_with_fal
+          ref: add-profiles-for-fal-ci  # change this to act1
```
chamini2, line 52: TODO

.github/workflows/test_dbt_fal.yml
```diff
@@ -0,0 +1,167 @@
...
+          # the existing revision.
+
+          pip install dbt-${{ matrix.profile }} -c fal/adapter/compatible_adapters.txt
+
+          pushd dbt-core
+          pip install -r dev-requirements.txt -r editable-requirements.txt
+          popd
+
+          pushd fal
+          pip install ".[${{ matrix.profile }}]"
```
chamini2, line 87: I think we don't need extras here, since we are taking care database read/write in the `fal/adapter` code.

pyproject.toml
```diff
@@ -43,7 +42,7 @@ firebase-admin = "^5.1.0"
 # Adapters
 
 ## snowflake
-snowflake-connector-python = { version = "2.7.9", optional = true } # 2.7.10 needs pyarrow 8
+snowflake-connector-python = { version = "<2.8", optional = true }
```
chamini2, line 45: if we really do not need the fal extras installed, we can revert this change

.github/workflows/test_dbt_fal.yml
isidentical: I thought it would do that by default  (e.g. only work on branches that inherit from `act1`)

chamini2: you mean because of the `push` instruction? that is isolated from this one

.github/workflows/test_dbt_fal.yml
```diff
@@ -0,0 +1,167 @@
...
+  run:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        profile: ["postgres", "bigquery", "snowflake", "duckdb"]
+        python: ["3.8"]
+        environment: ["local", "venv", "conda"]
+        # bigquery implementation fails due to concurrency, let's
+        # limit it to 1 for now.
```
chamini2, line 26: Thinking about it: this probably will fail for snowflake too. Want to limit it?

.github/workflows/test_dbt_fal.yml
chamini2: Well, I guess that is true.

.github/workflows/test_dbt_fal.yml
isidentical: It works for now, not sure what's going on but we can limit.

isidentical: I've tried without any concurrency, let's see whether it would work now (instead of running venv, local, conda parallel; now running them sequentially one by one in the same job)

adapter/pyproject.toml
```diff
@@ -12,9 +12,28 @@ python = ">=3.7.1, <3.11"
 dbt-core = "1.3.0b2"
 fal = ">=0.6.0"
 
+## snowflake
+snowflake-connector-python = { version = "^2.7.12", extras = ["pandas"], optional = true }
+
+## bigquery
+# We need the 1.5.0 version to support passing our own client.
```
chamini2, line 19: ATL: should it be 1.5 then? what are we doing about this?

isidentical: Just to keep it here I guess, but yeah it doesn't work out of the box. I tried to add a git dependency, but seems like poetry doesn't work when you mark something as both coming from git and optional.

isidentical:
Our solution seems to be not working for dbt-snowflake==1.0.x, needs investigation 🤔

PR #562, opened 2022-09-29
make database and schema no longer needed in credentials config

chamini2:
now this works:

```
  duckdb:
    type: duckdb
    path: "/Users/<user>/duck_db_dbt_dump.db" 
    python_adapter:
      type: fal
````

no more `database: None` or `schema: None`

<details>
closes fea-426
</details>

### DIFF
--- a/adapter/src/dbt/adapters/fal/connections.py
+++ b/adapter/src/dbt/adapters/fal/connections.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from typing import Tuple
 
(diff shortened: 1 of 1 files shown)


isidentical (approved):
Nice!

isidentical:
If we pass `database: None`, does that still work (or does it fail?)

chamini2:
> If we pass `database: None`, does that still work (or does it fail?)

It still works, we always just ignored it. And given that was yaml, we actually had a `"None"` string in there, not a `None` value.

PR #566, opened 2022-10-05
feat(fal-dbt): Teleport-based data transfer 

chamini2:
With the duckdb profile:

```yaml
    quack:
      type: duckdb
      path: "/Users/<user>/duck_db_dbt_dump.db" 
      python_adapter:
        type: fal
        teleport:
          type: local
          local_path: "/tmp/teleport"
````

It will load parquet files as necessary for transfer in the `/tmp/teleport` path.

Still have to test with S3, but the idea is there.

*****

This works with the `act1` branch of fal-ai/dbt-core.

### CODE REVIEW

**Comment by chamini2** [comment]
- duckdb dbt-fal adapter failing because `LOAD 'parquet'` is failing. Check https://duckdb.org/docs/extensions/overview
- check why bigquery is failing

File: adapter/src/dbt/adapters/fal/connections.py [hunk 1]
```diff
@@ -23,7 +44,9 @@ def cancel(self, connection):
 class FalCredentials(Credentials):
     default_environment: str = "local"
 
-    # NOTE: So we can not set them in profiles.yml
+    teleport: Optional[TeleportCredentials] = None
```
isidentical, line 47: ATL: does DBT allow nested credentials?

adapter/src/dbt/adapters/fal/teleport.py
```diff
@@ -0,0 +1,70 @@
...
+    locations: DataLocation,
+) -> AdapterResponse:
+    """Run the 'main' function inside the given code on the
+    specified environment.
+
+    The environment_name must be defined inside fal_project.yml file
+    in your project's root directory."""
+
+    with environment.connect() as connection:
+        execute_model = partial(run_with_teleport, code, teleport_info, locations)
```
isidentical, line 68: This is amazing that we can get away without sending the whole DBT config and manifest and doing all those hacks 💯 Perfection.

isidentical (approved):
This is quite amazing @chamini2, I've went through the code and beside a few stylistic comments (which guess we can ignore for now) I think it looks quite good.

adapter/src/dbt/adapters/fal/connections.py
chamini2: Well, it is not _really_ a credentials object in the same sense that the FalCredentials are. It's just the way to access/config Teleport; but in this case dbt just sees it as an object to parse.

PR #571, opened 2022-10-10
chore: remove pyarrow direct dependency

chamini2:
## Description

Remove pyarrow dependency and adjust versions so lockfile can be generated.

We had to upgrade bigquery because it expected `pyarrow < 8.dev`, but we are OK https://github.com/dbt-labs/dbt-bigquery/pull/332 for now. Still, we have to maybe look into markers for the future.

### Integration tests

Adapter to test:

- postgres
- snowflake
- bigquery
- redshift
- duckdb
- athena

Python version to test:

- 3.8
- 3.7
- 3.9
- 3.10

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -17,7 +17,7 @@ six = ">=1.9.0"
 
 [package.extras]
 docs = ["Sphinx (>=1.2.2)", "sphinx-rtd-theme (>=0.1.6)"]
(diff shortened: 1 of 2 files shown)


chamini2:
Failing when recreating the environment: https://github.com/fal-ai/fal/actions/runs/3221520624/jobs/5269566639#step:8:7347

isidentical:
> I think the only solution is to do two separate pip calls.

@chamini2 but then we would end up overriding the `dbt-bigquery`'s dependency requirements, no? (or does `dbt-bigquery` just work with `google-bigquery-client==3.3.3`?).

chamini2:
> > I think the only solution is to do two separate pip calls.
> 
> @chamini2 but then we would end up overriding the `dbt-bigquery`'s dependency requirements, no? (or does `dbt-bigquery` just work with `google-bigquery-client==3.3.3`?).

It does work https://github.com/dbt-labs/dbt-bigquery/issues/269

PR #573, opened 2022-10-11
chore: remove --keyword flag

mederka:
`fal` is the only keyword flag that we will support

## Description
Drop support for the unnecessary `--keyword` flag

### Integration tests

Adapter to test:
- postgres

Python version to test:
- 3.8


chamini2 (approved):
Thanks!

PR #574, opened 2022-10-11
feat: add --globals flag for more explicit global script execution

mederka:
## Description
Add `--globals` flag for more explicit global script execution. With this change, if any model selector flag OR a `--scripts` flag are passed, global scripts are NOT executed by default. In order to run them in these scenarios, users need to pass the `--globals` flag.

Following scenarios are now tested in integrations tests:

- fal runs globals by default when no model or script selection is indicated
- fal run doesn't trigger globals with select flag
- fal run triggers global scripts with model select and `--globals` flags
- script selection with `--scripts` flag doesn't run global scripts
- global scripts are run when both `--scripts` and `--globals` flags are passed

### Integration tests

Adapter to test:
- postgres

Python version to test:
- 3.8

### DIFF
--- a/integration_tests/features/globals.feature
+++ b/integration_tests/features/globals.feature
@@ -12,11 +12,20 @@ Feature: global scripts
     Then the following scripts are ran:
       | some_model.after.py | GLOBAL.after.py |
 
-  Scenario: fal run triggers globals with select flag
+  Scenario: fal run doesn't trigger globals with select flag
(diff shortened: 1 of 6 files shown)


chamini2 (approved):
Thank you! This is a breaking change, so we will ned to do a minor bump when releasing.

chamini2:
We should add it to telemetry!

PR #578, opened 2022-10-12
chore: introduce fal-models-paths variable for fal Python Models

chamini2:
## Description

If we are on dbt 1.3, `.py` files inside of the `models` directory are considered models already.

### Integration tests

Adapter to test:

- postgres
- snowflake
- bigquery
- redshift
- duckdb
- athena

Python version to test:

- 3.8


src/faldbt/parse.py
```diff
@@ -20,6 +21,7 @@
     from fal.packages.environments import BaseEnvironment
 
 FAL_SCRIPTS_PATH = "fal-scripts-path"
+FAL_MODELS_PATHS = "fal-models-paths"
```
isidentical, line 24: ATL: why are we supporting multiple paths for model directories?

src/fal/utils.py
```diff
@@ -59,10 +59,10 @@ def cache_static(func):
     _function_cache = _NO_HIT
 
     @functools.wraps(func)
-    def wrapper():
+    def wrapper(*func_args, **func_kwargs):
```
isidentical, line 62: Can you also change the docstring to reflect this?

src/fal/cli/model_generator/model_generator.py
```diff
@@ -103,21 +114,44 @@ def _check_path_safe_to_write(sql_path: Path, py_path: Path):
...
+        yield from func(models_path)
+
+
+def _find_fal_python_models_and_sql_target(base_sql_path: Path, models_path: Path):
+    for py_path in _find_python_files(models_path):
+        sql_path = _sql_path_from_python_path(
+            base_sql_path, py_path.relative_to(models_path)
+        )
+        old_cecksum = _check_path_safe_to_write(sql_path, py_path)
+        yield py_path, sql_path, old_cecksum
```
isidentical, line 139: NIT: `old_checksum`

isidentical (approved):
Wow, thanks for making the PR so quickly @chamini2. Only a couple of questions / nits, but the overall logic and implementation looks great 🚢

src/faldbt/parse.py
isidentical: Ah, didn't know that. Thanks for the explanation!

src/fal/utils.py
chamini2: actually changed it to import a backported version

src/faldbt/parse.py
```diff
@@ -81,22 +83,44 @@ def get_el_configs(
...
+    return {**vars, **cli_vars}
+
+
+@cache_static
+def get_fal_models_dirs(project_dir: str, args_vars: str) -> List[str]:
+    vars = get_vars_dict(project_dir, args_vars)
+    model_paths = vars.get(FAL_MODELS_PATHS) or []
+    if not model_paths:
+        # None or empty list
+        LOGGER.warn(
```
turbo1912, line 103: we should maybe log this to telemetry

PR #579, opened 2022-10-13
fix: make teleport work with s3

mederka:
Get teleport to work with S3

### DIFF
--- a/adapter/src/dbt/adapters/fal/teleport.py
+++ b/adapter/src/dbt/adapters/fal/teleport.py
@@ -1,6 +1,7 @@
 from typing import Any, Callable, Dict, NewType
 from functools import partial
 import functools
+from dbt.adapters.fal.connections import TeleportTypeEnum
 import pandas as pd
 
 from dbt.contracts.connection import AdapterResponse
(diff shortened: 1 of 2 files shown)


adapter/src/dbt/adapters/fal/teleport.py
```diff
@@ -37,12 +39,23 @@ def _teleport_df_to_external_storage(teleport_info: TeleportInfo, locations: Dat
...
         raise RuntimeError(f"Format {teleport_info.format} not supported")
 
+def _build_teleport_storage_options(teleport_info: TeleportInfo) -> Dict[str, str]:
+    storage_options = {}
+    if teleport_info.credentials.type == TeleportTypeEnum.REMOTE_S3:
+        storage_options = {
+            "key": teleport_info.credentials.s3_access_key_id,
+            "secret": teleport_info.credentials.s3_access_key
+        }
+    return storage_options
```
chamini2, line 57: Let's do 

```py
elif ... == LOCAL:
  pass
else:
  raise ...
```

So we get an exception if we forget to add a type here

chamini2 (approved):
This is very good! Small change

chamini2 (approved):
Awesome!

PR #584, opened 2022-10-14
feat: teleport between snowflake and S3

mederka:
## Description
Makes teleport between Snowflake and S3 possible

### Integration tests

Python version to test:
- 3.8

### DIFF
--- a/adapter/src/dbt/adapters/fal/teleport_support/snowflake.py
+++ b/adapter/src/dbt/adapters/fal/teleport_support/snowflake.py
@@ -1,0 +1,66 @@
+from typing import List
+from dbt.adapters.fal.connections import TeleportCredentials
+from dbt.fal.adapters.teleport.impl import TeleportAdapter
+from dbt.fal.adapters.teleport.info import TeleportInfo
+from dbt.adapters.fal.adapter_support import new_connection
+from dbt.adapters.base.relation import BaseRelation
+from dbt.adapters.base.impl import BaseAdapter
+
+
+class SnowflakeAdapterTeleport(TeleportAdapter):
+
+    def __init__(self, db_adapter: BaseAdapter, teleport_credentials: TeleportCredentials):
+        self._db_adapter = db_adapter
+        self._credentials = teleport_credentials
+        url = f's3://{teleport_credentials.s3_bucket}/teleport'
+
+        from dbt.adapters.fal.adapter_support import new_connection
+
+        with new_connection(self._db_adapter, "fal-snowflake:setup-teleport") as conn:
+            cur = conn.handle.cursor()
+
+            create_stage_query = f"""CREATE OR REPLACE STAGE falstage
+            URL = '{url}' CREDENTIALS = (
+            aws_key_id='{self._credentials.s3_access_key_id}',
+            aws_secret_key='{self._credentials.s3_access_key}');"""
+
+            create_format_query = """CREATE OR REPLACE FILE FORMAT falparquet type = 'PARQUET';"""
+
+            cur.execute(create_stage_query)
+            cur.execute(create_format_query)
+
+
+    @classmethod
+    def storage_formats(cls):
+        return ['parquet']
+
+    def teleport_from_external_storage(self, relation: BaseRelation, relation_path: str, teleport_info: TeleportInfo, columns: List[str]) -> None:
+        assert teleport_info.format == 'parquet', "snowflake only supports parquet format for Teleport"
+        location = f"@falstage/{str(relation).lower()}.parquet"
+        columns_str = ', '.join([f"$1:{c}" for c in columns])
+
+        rendered_macro = self._db_adapter.execute_macro(
+            'snowflake__copy_from',
+            kwargs={
+                'relation': relation,
+                'location': location,
+                'columns': columns_str
+            })
+
+        with self._db_adapter.connection_named('teleport:copy_to'):
+            self._db_adapter.execute(rendered_macro)
+
+    def teleport_to_external_storage(self, relation: BaseRelation, teleport_info: TeleportInfo) -> str:
+        assert teleport_info.format == 'parquet', "snowflake only supports parquet format for Teleport"
+        relation_path = teleport_info.build_relation_path(relation)
+        location = f"@falstage/{str(relation).lower()}.parquet"
+        rendered_macro = self._db_adapter.execute_macro(
+            'snowflake__copy_to',
+            kwargs={
+                'relation': relation,
+                'location': location
+            })
+        with self._db_adapter.connection_named('teleport:copy_from'):
+            self._db_adapter.execute(rendered_macro)
+        return relation_path
+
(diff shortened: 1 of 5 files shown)


adapter/src/dbt/adapters/fal/teleport_support/snowflake.py
```diff
@@ -0,0 +1,66 @@
...
+        self._db_adapter = db_adapter
+        self._credentials = teleport_credentials
+        url = f's3://{teleport_credentials.s3_bucket}/teleport'
+
+        from dbt.adapters.fal.adapter_support import new_connection
+
+        with new_connection(self._db_adapter, "fal-snowflake:setup-teleport") as conn:
+            cur = conn.handle.cursor()
+
+            create_stage_query = f"""CREATE OR REPLACE STAGE falstage
```
mederka, line 22: Couldn't put this in a macro

adapter/src/dbt/adapters/fal/teleport_support/snowflake.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+    @classmethod
+    def storage_formats(cls):
+        return ['parquet']
+
+    def teleport_from_external_storage(self, relation: BaseRelation, relation_path: str, teleport_info: TeleportInfo, columns: List[str]) -> None:
+        assert teleport_info.format == 'parquet', "snowflake only supports parquet format for Teleport"
+        location = f"@falstage/{str(relation).lower()}.parquet"
+        columns_str = ', '.join([f"$1:{c}" for c in columns])
```
mederka, line 40: Should the column name be quoted? As in `f"$1:'{c}'"`.

adapter/src/dbt/adapters/fal/teleport_support/snowflake.py
```diff
@@ -0,0 +1,66 @@
...
+                'columns': columns_str
+            })
+
+        with self._db_adapter.connection_named('teleport:copy_to'):
+            self._db_adapter.execute(rendered_macro)
+
+    def teleport_to_external_storage(self, relation: BaseRelation, teleport_info: TeleportInfo) -> str:
+        assert teleport_info.format == 'parquet', "snowflake only supports parquet format for Teleport"
+        relation_path = teleport_info.build_relation_path(relation)
+        location = f"@falstage/{str(relation).lower()}.parquet"
```
mederka, line 56: So I'm not sure if we should have a dedicated function for constructing this string.

adapter/src/dbt/adapters/fal/teleport_support/snowflake.py
chamini2: Super curious about why we couldnt. It would not render it? The data passed have errors?

adapter/src/dbt/adapters/fal/teleport_support/snowflake.py
chamini2: > `'$1:a, $1:b, $1:c'`

This is the desired output? With the `$1`?

adapter/src/dbt/adapters/fal/teleport_support/snowflake.py
chamini2, line 56:
```suggestion
        location = f"@falstage/{relation_path}"
```

adapter/src/dbt/include/fal/macros/teleport_snowflake.sql
```diff
@@ -0,0 +1,21 @@
...
+    CREATE OR REPLACE TABLE {{ relation }} USING template (
+      SELECT array_agg(object_construct(*))
+	  	FROM table(
+  			infer_schema(
+	  			LOCATION=>'{{ location }}',
+		  		file_format=>'falparquet'
+			  )
+  		)
+    );
+    COPY INTO {{ relation }} FROM (SELECT {{ columns }} FROM {{ location }})
```
chamini2, line 11: I think we could maybe break this into 2 macros?

1. Macro with `CREATE OR REPLEACE TABLE...`
2. Fetch the newly created table's column names
  a. could be `DESCRIBE {{ relation }}`
  b. could be a dbt method `adatper.fetch_columns` or maybe reloading the relation info will bring the columns?
  c. could be `SELECT column_name FROM {{ relation.database }}.information_schema.columns WHERE table_name ilike '{{ relation.identifier }}';`
3. Macro with `COPY INTO` that get's the columns passed
  a. A comment: you could pass the `columns` list, and do a for-loop inside Jinja

adapter/src/dbt/adapters/fal/teleport_support/snowflake.py
```diff
@@ -0,0 +1,66 @@
...
+            cur.execute(create_format_query)
+
+
+    @classmethod
+    def storage_formats(cls):
+        return ['parquet']
+
+    def teleport_from_external_storage(self, relation: BaseRelation, relation_path: str, teleport_info: TeleportInfo, columns: List[str]) -> None:
+        assert teleport_info.format == 'parquet', "snowflake only supports parquet format for Teleport"
+        location = f"@falstage/{str(relation).lower()}.parquet"
```
chamini2, line 39:
```suggestion
        location = f"@falstage/{teleport_info.build_relation_path(relation)}"
```

adapter/src/dbt/adapters/fal/teleport_support/snowflake.py
mederka: Yes, by default it creates a table with 1 column that has {'key':'value'} dicts.

adapter/src/dbt/adapters/fal/teleport_support/snowflake.py
mederka: It would not render it and just get stuck. This is before any logs start printing

adapter/src/dbt/adapters/fal/teleport_support/snowflake.py
mederka: https://docs.snowflake.com/en/user-guide/script-data-load-transform-parquet.html#sql-script-1-load-parquet-data

adapter/src/dbt/include/fal/macros/teleport_snowflake.sql
```diff
@@ -0,0 +1,30 @@
...
+    );
+{%- endmacro %}
+
+{% macro snowflake__copy_from_parquet(relation, location) -%}
+    COPY INTO {{ relation }} FROM (
+      SELECT
+        {% for col in adapter.get_columns_in_relation(relation) %}
+          $1:{{col.column}}
+          {%- if not loop.last -%},{%- endif -%}
+        {% endfor %}
```
chamini2, line 19: This is great!

adapter/src/dbt/adapters/fal/teleport_support/snowflake.py
```diff
@@ -0,0 +1,73 @@
...
+from dbt.adapters.base.relation import BaseRelation
+from dbt.adapters.base.impl import BaseAdapter
+
+
+class SnowflakeAdapterTeleport(TeleportAdapter):
+
+    def __init__(self, db_adapter: BaseAdapter, teleport_credentials: TeleportCredentials):
+        self._db_adapter = db_adapter
+        self._credentials = teleport_credentials
+        url = f's3://{teleport_credentials.s3_bucket}/teleport'
```
chamini2, line 15: Should this be handled by a `TeleportInfo` object or something similar?

https://github.com/fal-ai/fal/blob/main/adapter/src/dbt/fal/adapters/teleport/info.py#L72-L73

*****

We can solve this later, but maybe add a TODO?

chamini2 (approved):
I think we are good!

PR #589, opened 2022-10-18
make encapsulate work with isolated environments

isidentical:


adapter/src/dbt/adapters/fal/utils.py
```diff
@@ -65,3 +68,16 @@ def fetch_environment(project_root: str, environment_name: str) -> Tuple[BaseEnv
...
     return environments[environment_name], False
+
+
+def db_adapter_config(config: RuntimeConfig) -> RuntimeConfig:
+    """Return a config object that has the database adapter as its primary. Only
+    applicable when the underlying db adapter is encapsulated."""
+    if hasattr(config, "sql_adapter_credentials"):
+        python_adapter_credentials = config.credentials
+        new_config = replace(config, credentials=config.sql_adapter_credentials)
+        new_config.python_adapter_credentials = config.credentials
```
chamini2, line 79: TIL about replace, but does this make sense:

```suggestion
        new_config = replace(config, credentials=config.sql_adapter_credentials, python_adapter_credentials=config.credentials)
```

?

adapter/src/dbt/adapters/fal/utils.py
```diff
@@ -65,3 +68,16 @@ def fetch_environment(project_root: str, environment_name: str) -> Tuple[BaseEnv
         )
 
     return environments[environment_name], False
+
+
+def db_adapter_config(config: RuntimeConfig) -> RuntimeConfig:
+    """Return a config object that has the database adapter as its primary. Only
+    applicable when the underlying db adapter is encapsulated."""
```
chamini2, line 75: great approach

chamini2 (approved):
Thanks for getting this to work! Check a few coments

isidentical:
This should be good to go, thanks guys I'll try to look into the reviews @chamini2 @mederka

turbo1912 (approved):

adapter/src/dbt/adapters/fal/utils.py
isidentical: AFAIK `python_adapter_credentials` is just an attribute we attach at the runtime, not something that is on the config (might be wrong on this) of the official dbt-core==1.3. `replace()` only works for standard parameters:
```
>>> from dataclasses import dataclass, replace
>>> 
>>> @dataclass
... class T:
...     name: str
... 
>>> 
>>> t = T(name="hey")
>>> replace(t, name="x")
T(name='x')
>>> replace(t, name="x", random_attr="y")
  File "<stdin>", line 1, in <module>
  File "/home/seaworth/.pyenv/versions/3.9.13/lib/python3.9/dataclasses.py", line 1284, in replace
    return obj.__class__(**changes)
TypeError: __init__() got an unexpected keyword argument 'random_attr'
```

PR #594, opened 2022-10-28
Disable inherit_local

isidentical:
- Disables `inherit_local` by default. This was an hack we have used to make it work with act1 branches of other packages; but we shouldn't need it now.
- Drops the redundant `fal_adapter_support.py` which doesn't look like used anywhere and is leftover from the initial version of the dbt-encapsulate adapter.
- Installs `dbt-fal` from PyPI as if it was just a regular adapter, but with all the required extras that are actually present in the local environment.


chamini2 (approved):
Thanks!

PR #601, opened 2022-10-31
fix: ref function for duckdb in encapsulated adapter

mederka:


chamini2 (approved):
What happend if not?

mederka:
> What happend if not?

Couldn't find the table

PR #604, opened 2022-11-01
feat: make encapsulate pass-through sql tables

chamini2:
## Description

With this change now Python models with encapsulating adapter can have 

```py
def model(dbt, fal):
  dbt.config(materialized='table')
```

and it will run on fal.

❗ ❗❗ This makes it impossible to run some Python models on fal and others on the db-adapter's Python runtime. All models will run with fal.

### Integration tests

- [x] We have to wait for https://github.com/fal-ai/fal/pull/602 to be merged first.

Adapter to test:

- postgres
- snowflake
- bigquery
- duckdb

Python version to test:

- 3.8

### DIFF
--- a/adapter/src/dbt/adapters/fal/wrappers.py
+++ b/adapter/src/dbt/adapters/fal/wrappers.py
@@ -9,6 +9,7 @@ from dbt.clients.jinja import MacroGenerator
 
 from dbt.adapters.fal_experimental.impl import FalAdapterMixin
 from fal.telemetry import telemetry
+from fal.utils import cache_static
 
 class FalCredentialsWrapper:
     _db_creds: Optional[Credentials] = None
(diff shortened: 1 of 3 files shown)


adapter/src/dbt/adapters/fal/wrappers.py
```diff
@@ -41,6 +45,18 @@ def __init__(self, db_adapter_type: Type[BaseAdapter], config):
     def submit_python_job(self, *args, **kwargs):
         return super().submit_python_job(*args, **kwargs)
 
+    @available
+    def db_materialization(self, context, materialization):
+        # NOTE: inspired by https://github.com/dbt-labs/dbt-core/blob/be4a91a0fe35a619587b7a0145e190690e3771c6/core/dbt/task/run.py#L254-L290
+        manifest = ManifestLoader.get_full_manifest(self.config)
```
isidentical, line 51: ATL (or a future TODO): should we consider caching the manifests? E.g. if you have tons of models, we would be generating a new manifest for each one of them (but IIRC one should suffice).

isidentical (approved):
Super clean and straighforward, thanks a lot for looking into this @chamini2 🚀 One question below (but it doesn't affect this PR, maybe a future item if you agree)

PR #608, opened 2022-11-01
feat(dbt-fal): dbt-fal with remote isolate

mederka:


adapter/tools/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# For building this image, you need to be in the project
```
turbo1912, line 1: not sure if we need to put this dockerfile here

adapter/src/dbt/adapters/fal_experimental/utils.py
```diff
@@ -109,21 +109,32 @@ def load_environments(base_dir: str) -> Dict[str, "BaseEnvironment"]:
...
         raise ValueError(
             f"Invalid environment type (of {kind}) for {name}. Please choose from: "
             + ", ".join(REGISTERED_ENVIRONMENTS.keys())
         )
 
     parsed_config = {
-        'requirements': config.get('requirements', [])
-    }
+            'requirements': config.get('requirements', [])
+        }
```
turbo1912, line 130: nothing changed here

adapter/src/dbt/adapters/fal_experimental/utils.py
```diff
@@ -109,21 +109,32 @@ def load_environments(base_dir: str) -> Dict[str, "BaseEnvironment"]:
...
+
+    if kind == "remote":
```
turbo1912, line 132: how about adding a host to the profiles.yml file? we should enable remote mode through that as well

```
jaffle_shop:
  target: fal
  outputs:
    dev:
      type: duckdb
      path: "/Users/gorkemyurtseven/dev/duck_db_dbt_dump"
    fal:
      type: fal
      db_profile: dev
      host: http://146.83.102
```

isidentical: This was an open discussion point. We can either support per-environment level remotes OR we can support whole remote and whole local experience (by whether `host` is present in the credentials/config or not).

chamini2: I think it's easier to work with and think about if we do by project instead of by environment.

adapter/tools/Dockerfile
mederka: This was an easy way to get an image with dbt-fal in it.

adapter/src/dbt/adapters/fal_experimental/utils/environments.py
```diff
@@ -131,6 +135,25 @@ def _get_required_key(data: Dict[str, Any], name: str) -> Any:
...
+    remote_type = REMOTE_TYPES_DICT.get(config["remote_type"])
+
+    assert remote_type, f"{config['remote_type']} not recognised. Available remote types: {list(REMOTE_TYPES_DICT.keys())}"
+
+    return {
+        "host": config.get("host"),
+        "target_environment_kind": remote_type,
+        "target_environment_config": {
+            'requirements': config.get('requirements', []),
+        }
```
isidentical, line 155: If it were to be conda, then this has to be packages. I think we can simply use all the remaining parameters from the config here (except kind / host etc.)

adapter/src/dbt/adapters/fal_experimental/adapter.py
```diff
@@ -29,29 +31,46 @@ def run_with_adapter(code: str, adapter: BaseAdapter) -> Any:
...
+    from isolate.backends.remote import IsolateServer
     """Run the 'main' function inside the given code on the
     specified environment.
 
     The environment_name must be defined inside fal_project.yml file
     in your project's root directory."""
+    deps = [i for i in get_default_pip_dependencies() if i.startswith('dbt-')]
+    if type(environment) == IsolateServer:
+        if environment.target_environment_kind == 'conda':
+            environment.target_environment_config['packages'].extend(deps)
```
isidentical, line 57: I am still not sure about this, considering `deps` are pip packages and not conda (e.g.`x==y` vs `x=y`, plus there might be packages that doesn't exist on the conda forge).

But it also means we can't solve it for variadic dbt dependencies now, since we can't create multiple environments. So maybe, now is the time to actually implement inheritance among environments? (In any case, for this PR, I'd recommend just raising an error when the target is conda on remote; since there is no easy fix).

isidentical (approved):

isidentical:
Looks great, amazing work @mederka!

PR #612, opened 2022-11-02
chore(dbt-fal): add dbt-fal publishing

chamini2:
## Description
This is designed to publish each package at a time.

One problem is changelog generation will be weird. We probably will need to start separating commits like:

- `fix(fal): ...`: fal-cli only
- `fix(dbt-fal): ...` or `fix(adapter): ...`: adapter only
- `fix: ...` all


chamini2:
image

turbo1912 (approved):

.github/workflows/python-release.yml
```diff
@@ -5,6 +5,14 @@ name: PyPI Release
...
   workflow_dispatch:
     inputs:
+      package:
+        description: Package
+        required: false
+        default: fal
+        type: choice
+        options:
+          - fal
+          - dbt-fal
```
isidentical, line 15: This is super cool (being able to choose which to publish)

isidentical (approved):
Agree on the changelog point.

.github/workflows/python-release.yml
```diff
@@ -35,11 +48,23 @@ jobs:
...
 
+      - name: Package setup
+        run: |
+          if [[ ${{ inputs.package }} == fal ]]
+          then
+            "PACKAGE_DIR=." >> $GITHUB_ENV
+            "TAG_PREFIX=v" >> $GITHUB_ENV
+          else if [[ ${{ inputs.package }} == dbt-fal ]]
+            "PACKAGE_DIR=adapter" >> $GITHUB_ENV
+            "TAG_PREFIX=adapter-v" >> $GITHUB_ENV
```
isidentical, line 59: This is super clever btw 💯

isidentical:
@chamini2 git cliff seems to have a path filter option

```
    --include-path <PATTERN>...    Sets the path to include related commits [env: GIT_CLIFF_INCLUDE_PATH=]
    --exclude-path <PATTERN>...    Sets the path to exclude related commits [env: GIT_CLIFF_EXCLUDE_PATH=]
```

I wonder, at least at the beginning, whether it is easier to configure in terms of paths instead of commit messages?

PR #632, opened 2022-11-04
fix(dbt-fal): Find wrapped adapter Relation class on init 

chamini2:
We no longer have to specify the wrapped adapter's quoting behavior in dbt_project.yml


isidentical (approved):
Wow 🤯 Super cool work!

PR #633, opened 2022-11-04
fix(dbt-fal): Inherit is_cancelable from the wrapped adapter

chamini2:
## Description

Now when a `ctrl-c` happens, it cancels or not depending on the wrapped db adapter config.

We may need to work on handling cancels better in the future. But it's OK for now.


isidentical (commented):
Not super sure why we are moving this @chamini2, isn't `FalAdapterMixin`'s `is_cancallable` already inherited in `FalAdapter` (or is there something weird going on 🤔)?

chamini2:
`FalAdapter` extends `FalAdapterMixin`, so for the `FalEncAdapterWrapper` it will use the DB's implementation.

isidentical:
Ah, ok. Makes sense

isidentical (approved):

PR #643, opened 2022-11-08
fix(dbt-fal): make dbt-fal work with redshift

mederka:
- Fixes installation of dbt-fal with `pip install dbt-fal[redshift]`
- Adds redshift to integration tests

### DIFF
--- a/.github/workflows/test_integration_adapter.yml
+++ b/.github/workflows/test_integration_adapter.yml
@@ -30,6 +30,7 @@ jobs:
           - bigquery
           - duckdb
           - snowflake
+          - redshift
         python: ["3.7", "3.8"]
         include:
           - profile: duckdb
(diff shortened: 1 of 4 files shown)


.github/workflows/test_integration_adapter.yml
```diff
@@ -64,6 +65,11 @@ jobs:
             pip install 'dbt-${{ matrix.profile }}==${{ matrix.adapter_version }}'
           fi
 
+          if [[ '${{ matrix.profile }}' == "redshift" ]]
+          then
+            pip install dbt-postgres
+          fi
+
```
chamini2, line 72: I think we don't have to install it, it's a dependency: https://github.com/dbt-labs/dbt-redshift/blob/main/setup.py#L70

adapter/pyproject.toml
```diff
@@ -46,6 +48,7 @@ bigquery = ["google-cloud-bigquery"]
 redshift = ["sqlalchemy-redshift"]
 duckdb = []
 athena = []
+teleport = ["s3fs"]
```
chamini2, line 51: Can we make a ticket to

```py
yield 'dbt-fal[<adapter>,teleport]', ...
```

when using teleport?

mederka: We have a "test teleport" ticket, and I'm sure we will resolve this issue there.

.github/workflows/test_integration_adapter.yml
mederka: Good catch!

PR #664, opened 2022-11-14
chore(dbt-fal): enable isolate server multiple envs

mederka:


adapter/src/dbt/adapters/fal_experimental/adapter.py
```diff
@@ -57,10 +57,12 @@ def run_in_environment_with_adapter(
...
-        if environment.target_environment_kind == 'conda':
-            raise NotImplementedError("Remote environment with `conda` is not supported yet.")
-        else:
-            environment.target_environment_config['requirements'].extend(deps)
+        extra_config = {
+            'kind': 'virtualenv',
+            'configuration': { 'requirements': deps }
+        }
+
+        environment.target_environments.append(extra_config)
```
chamini2, line 65: ATL: does this now support conda? Does `environment.target_environments` have 1 or more items?

mederka: This will allow us to cache dbt-* packages and speed up env setup. Not sure about conda, @isidentical.

isidentical (approved):
Amazing work @mederka both on the server implementation and the client here!

PR #669, opened 2022-11-16
Fix docs for post-hook

burkaygur:
### Integration tests

Adapter to test:

- postgres

Python version to test:

- 3.8


docsite/docs/fal/reference/fal-scripts-path.md
```diff
@@ -52,11 +53,12 @@ models:
           before:
             # searches in `./` because var has default value
             - scripts/before.py
-          post-hook:
-            - scripts/after.py
+        post-hook:
+          - scripts/after.py
```
chamini2, line 57: good catch

chamini2 (approved):
thanks!

PR #670, opened 2022-11-17
chore: change default fal-scripts-path to fal_scripts

mederka:


mederka:
@chamini2, I'm actually not sure about this as it could be a breaking change for a lot of our users. Should we make project root the default scripts directory in dbt-fal instead?

PR #671, opened 2022-11-17
chore(dbt-fal): change default script path to project dir

mederka:
This is done for consistency with fal. There might be future use cases where scripts path cannot be the dbt project path, so we have to be mindful of that.

### DIFF
--- a/adapter/integration_tests/projects/simple_project/dbt_project.yml
+++ b/adapter/integration_tests/projects/simple_project/dbt_project.yml
@@ -12,6 +12,3 @@ target-path: "{{ env_var('temp_dir') }}/target"
 
 models:
   +schema: custom
-
-vars:
-  fal-scripts-path: "utils"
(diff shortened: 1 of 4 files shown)


adapter/integration_tests/projects/simple_project/dbt_project.yml
```diff
@@ -10,6 +10,9 @@ snapshot-paths: ["snapshots"]
 
 target-path: "{{ env_var('temp_dir') }}/target"
 
+vars:
+  fal-scripts-paths: "fal_scripts"
+
```
chamini2, line 15: We were already setting this in line 19 of this file. Which I guess overwrites this value and that is why it works.

chamini2 (commented):
The code change is 👍. However, can we make it that the one project of tests has default `fal-scripts-paths` and the other has it set in dbt_project.yml? This way we test it implicitly.

PR #673, opened 2022-11-18
fix(dbt-fal): trino none credential handling

mederka:
The user is now able to install dbt-fal[trino] and run it even if the older version of dbt-trino is installed.

### DIFF
--- a/adapter/src/dbt/adapters/fal_experimental/support/trino.py
+++ b/adapter/src/dbt/adapters/fal_experimental/support/trino.py
@@ -1,7 +1,7 @@
-from typing import Any, Dict, Optional
+from typing import Any, Dict
 from dbt.adapters.base import BaseAdapter, Credentials
 from trino.sqlalchemy import URL
-from dbt.adapters.trino.connections import TrinoCredentials, TrinoNoneCredentials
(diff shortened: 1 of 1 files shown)


adapter/src/dbt/adapters/fal_experimental/support/trino.py
```diff
@@ -17,10 +17,7 @@ def create_engine(adapter: BaseAdapter) -> Any:
     )
     return sqlalchemy.create_engine(url, connect_args=connect_args)
 
-def _build_connect_args(credentials: TrinoCredentials) -> Optional[Dict[str, Any]]:
-    if isinstance(credentials, TrinoNoneCredentials):
```
chamini2, line 21: ATL: why we no longer need to handle this case?

chamini2 (approved):

adapter/src/dbt/adapters/fal_experimental/support/trino.py
mederka: `TrinoNoneCredentials.trino_auth()` returns None and we need to pass http_scheme.
```
return {
        "auth": credentials.trino_auth(),
        "http_scheme": credentials.http_scheme.value
}
```

PR #676, opened 2022-11-22
Update README with dbt-fal

burkaygur:


chamini2 (approved):
This is a great change

PR #677, opened 2022-11-28
fix: FalDbt init with dbt-fal

mederka:
Initializing FalDbt in a directory other than project_dir was causing this error:

```
dbt.exceptions.DbtProjectError: Runtime Error
  no dbt_project.yml found at expected path [redacted]/dbt_project.yml
```


chamini2:
close in favor of #679 ? you change this solution in there

mederka:
Will be solved in #679

PR #679, opened 2022-11-29
tests: add dbt-fal to fal-cli integration tests

mederka:


src/faldbt/lib.py
```diff
@@ -338,6 +338,9 @@ def _write_relation(
     *,
     dtype=None,
 ) -> AdapterResponse:
+    if adapter.type() == "fal":
+        adapter = adapter._db_adapter
+
```
mederka, line 343: This addresses FEA-594

.github/actions/setup-local-fal/action.yml
```diff
@@ -51,8 +51,14 @@ runs:
           exit 2
         fi
 
-        echo "${package}==${version}"
-        pip install ${package}==${version}
+        if [[ '${{ inputs.adapter }}' == 'fal' ]]
+        then
+          pip install -e ./adapter
```
chamini2, line 56:
```suggestion
          pip install -e './adapter[postgres]'
```

chamini2: Also, should we install `dbt-fal` instead of the local version? 

PRO: not coupling them
CON: test latest changes together

.github/actions/setup-local-fal/action.yml
```diff
@@ -51,8 +51,14 @@ runs:
...
+          pip install dbt-postgres==${version}
+        else
+          echo "${package}==${version}"
+          pip install ${package}==${version}
+        fi
 
     - name: Install fal
       shell: sh
```
chamini2, line 64: Below this line there is a `pip install .[adapter]` which would be `.[fal]`.

src/faldbt/lib.py
chamini2: Awsome! Not necessary for fetch relations too?

src/faldbt/parse.py
```diff
@@ -72,8 +72,9 @@ def get_dbt_config(
         target=profile_target,
     )
 
-    if project_dir and not os.getenv("GITHUB_ACTIONS"):
+    if project_dir and not "PYTEST_CURRENT_TEST" in os.environ:
         # HACK: initializing dbt-fal requires cwd to be project_dir
+        # TODO: this doesn't work in pytest + Github Actions
```
chamini2, line 77: Could we consider somehow how to avoid the fact that

> initializing dbt-fal requires cwd to be project_dir


We could change the code in dbt-fal to check if it should cd into `project_dir` or understand where it is _standing_ right now better and not depend on that.

chamini2 (commented):
Thanks!

src/faldbt/lib.py
mederka: Yep!

.github/actions/setup-local-fal/action.yml
mederka: I think for now it's a good idea to keep them coupled. Maybe until 1.4?

.github/actions/setup-local-fal/action.yml
mederka: Good catch!

src/faldbt/parse.py
mederka: Yes! I looked into that and we would have to somehow communicate project-dir to the dbt-fal code that's causing the error. The error happens because of this line: https://github.com/fal-ai/fal/blob/main/adapter/src/dbt/adapters/fal/impl.py#L39 , which is executed when the module is being imported.

.github/actions/setup-local-fal/action.yml
```diff
@@ -63,7 +63,12 @@ runs:
...
       run: |
-        pip install -e '.[${{ inputs.adapter }}]'
+        if [[ '${{ inputs.adapter }}' == 'fal' ]]
+        then
+          pip install -e .
```
chamini2, line 68:
```suggestion
          pip install -e '.[postgres]'
```

chamini2 (approved):
LGTM! :shipit:

PR #680, opened 2022-11-29
tests: use latest 1.3 version for integration tests

chamini2:
Adapter to test:

- postgres
- snowflake
- bigquery
- redshift
- duckdb
- athena

Python version to test:

- 3.8
- 3.7
- 3.9
- 3.10


mederka (approved):
Looks good! We should create a ticket for fixing the adapter tests.

PR #683, opened 2022-12-07
feat(dbt-fal): enable fal cloud

mederka:
in `profiles.yml` you can now pass some credentials:
```
my_profile:
  target: fal_cloud
  outputs:
    fal_cloud:
      type: fal
      db_profile: db
      host: my_host
      secret: my_secret
      secret_id: my_secret_id
    db:
      type: postgres
      ...
```
And now, all Python models will run on isolate cloud by default. 

You can specify environments in `fal_project.yml`:
```
environments:
  - name: funny-cloud
    type: venv
    requirements:
      - pyjokes==0.6.0
```

You can specify machine sizes in model configs:
```
def model(dbt, fal):
    dbt.config(materialized="table")
    dbt.config(fal_machine="GPU") # "S" by default
    ...
```

If you want to run fal locally, you should omit credentials from `profiles.yml`:
```
my_profile:
  target: fal_local
  outputs:
    fal_local:
      type: fal
      db_profile: db
    db:
      type: postgres
      ...
```

### DIFF
--- a/.github/workflows/test_integration_adapter.yml
+++ b/.github/workflows/test_integration_adapter.yml
@@ -40,8 +40,6 @@ jobs:
           - profile: snowflake
             teleport: true
             cloud: true
-          - profile: bigquery
-            cloud: true
           - profile: redshift
             cloud: true
 
@@ -171,15 +169,21 @@ jobs:
           export DB_NAMESPACE="${{ github.run_id }}_${UUID}"
 
           BEHAVE_TAGS=""
-          if [[ '${{ matrix.teleport }}' != 'true' ]] && [[ '${{ matrix.cloud }}' != 'true' ]]
-          then
-            BEHAVE_TAGS="--tags=-teleport --tags=-cloud"
-          elif [[ '${{ matrix.teleport }}' != 'true' ]]
+          if [[ '${{ matrix.teleport }}' != 'true' ]]
           then
             BEHAVE_TAGS="--tags=-teleport"
-          elif [[ '${{ matrix.cloud }}' != 'true' ]]
+          fi
+
+          if [[ '${{ matrix.cloud }}' != 'true' ]]
           then
-            BEHAVE_TAGS="--tags=-cloud"
+            BEHAVE_TAGS="$BEHAVE_TAGS --tags=-cloud"
+          fi
+
+          if [[ -z "${GITHUB_HEAD_REF}" ]]
+          then
+            export FAL_GITHUB_BRANCH=${GITHUB_BASE_REF:-${GITHUB_REF#refs/heads/}}
+          else
+            export FAL_GITHUB_BRANCH=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}
           fi
 
           behave $BEHAVE_TAGS -fplain -D profile=${{ matrix.profile }}
--- a/.github/workflows/test_integration_adapter.yml
+++ b/.github/workflows/test_integration_adapter.yml
@@ -39,6 +39,12 @@ jobs:
             adapter_version: 1.2.2
           - profile: snowflake
             teleport: true
+            cloud: true
+          - profile: bigquery
+            cloud: true
+          - profile: redshift
+            cloud: true
+
 
     concurrency:
       group: "${{ github.head_ref || github.run_id }}-${{ github.workflow }}-${{ matrix.profile }}-${{ matrix.python }}"
@@ -75,13 +81,20 @@ jobs:
           fi
 
           pushd ..
-          if [[ '${{ matrix.teleport }}' == 'true' ]]
+          if [[ '${{ matrix.teleport }}' == 'true' ]] && [[ '${{ matrix.cloud }}' == 'true' ]]
+          then
+            pip install -e ".[${{ matrix.profile }},teleport,cloud]"
+          elif [[ '${{ matrix.teleport }}' == 'true' ]]
           then
             pip install -e ".[${{ matrix.profile }},teleport]"
+          elif [[ '${{ matrix.cloud }}' == 'true' ]]
+          then
+            pip install -e ".[${{ matrix.profile }},cloud]"
           else
             pip install -e ".[${{ matrix.profile }}]"
           fi
           popd
+
 
       - name: Setup behave
         working-directory: fal/adapter/integration_tests
@@ -116,6 +129,10 @@ jobs:
           # Teleport
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          # Isolate Cloud
+          FAL_HOST: ${{ secrets.FAL_HOST }}
+          FAL_KEY_SECRET: ${{ secrets.FAL_KEY_SECRET }}
+          FAL_KEY_ID: ${{ secrets.FAL_KEY_ID }}
         run: |
           source .venv/bin/activate
 
@@ -154,9 +171,15 @@ jobs:
           export DB_NAMESPACE="${{ github.run_id }}_${UUID}"
 
           BEHAVE_TAGS=""
-          if [[ '${{ matrix.teleport }}' != 'true' ]]
+          if [[ '${{ matrix.teleport }}' != 'true' ]] && [[ '${{ matrix.cloud }}' != 'true' ]]
+          then
+            BEHAVE_TAGS="--tags=-teleport --tags=-cloud"
+          elif [[ '${{ matrix.teleport }}' != 'true' ]]
           then
             BEHAVE_TAGS="--tags=-teleport"
+          elif [[ '${{ matrix.cloud }}' != 'true' ]]
+          then
+            BEHAVE_TAGS="--tags=-cloud"
           fi
 
           behave $BEHAVE_TAGS -fplain -D profile=${{ matrix.profile }}
--- a/adapter/src/dbt/adapters/fal_experimental/adapter.py
+++ b/adapter/src/dbt/adapters/fal_experimental/adapter.py
@@ -12,8 +12,9 @@ from dbt.contracts.connection import AdapterResponse
 
 from isolate.backends.virtualenv import PythonIPC, VirtualPythonEnvironment
 from dbt.adapters.fal_experimental.utils.environments import (
-    get_default_pip_dependencies,
+    get_default_pip_dependencies, _find_adapter_extras
 )
+
 from dbt.parser.manifest import MacroManifest, Manifest
 
 from isolate.backends import BaseEnvironment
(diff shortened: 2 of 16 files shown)


adapter/src/dbt/adapters/fal_experimental/adapter.py
```diff
@@ -47,16 +47,25 @@ def run_in_environment_with_adapter(
...
 
     The environment_name must be defined inside fal_project.yml file
     in your project's root directory."""
-    if type(environment) == IsolateServer:
-        # TODO: make a specialized function for this case?
+
+    if type(environment).__name__ in ['IsolateServer', 'FalHostedServer']:
         deps = [i for i in get_default_pip_dependencies() if i.startswith('dbt-')]
 
+        if not any('dbt-fal' in i for i in deps):
```
chamini2, line 59: in `get_default_pip_dependencies` there is already a special treatment of `dbt-fal`. Can we solve the hack in there instead of bringing it here?

adapter/src/dbt/adapters/fal_experimental/connections.py
```diff
@@ -43,8 +43,11 @@ def cancel(self, connection):
 @dataclass
 class FalCredentials(Credentials):
     default_environment: str = "local"
-
     teleport: Optional[TeleportCredentials] = None
+    host: str = ''
+    key_secret: str = ''
+    key_id: str = ''
```
chamini2, line 49: Maybe do we want to do an object in a single property instead of 3 separate properties so we know they are ALL in there. 

Something like
```
my_profile:
  target: fal_cloud
  outputs:
    fal_cloud:
      type: fal
      db_profile: db
      cloud: 
        host: my_host
        key_id: my_id
        key_secret: my_secret
    db:
      type: postgres
```

adapter/src/dbt/adapters/fal_experimental/impl.py
```diff
@@ -72,8 +72,16 @@ def submit_python_job(
             self.credentials.default_environment,
         )
 
+        machine_type = parsed_model["config"].get(
+            "fal_machine",
+            "S",
+        )
```
chamini2, line 78: Great!

adapter/src/dbt/adapters/fal_experimental/utils/environments.py
```diff
@@ -31,17 +35,27 @@ def run(self, executable: BasicCallable, *args, **kwargs) -> Any:
...
+        credentials: Optional[Any] = None
 ) -> Tuple[BaseEnvironment, bool]:
     """Fetch the environment with the given name from the project's
     fal_project.yml file."""
     # Local is a special environment where it doesn't need to be defined
     # since it will mirror user's execution context directly.
     if environment_name == "local":
-        return LocalPythonEnvironment(), True
+        if credentials.key_secret and credentials.key_id:
+            # "local" environment in this case is fal cloud
```
chamini2, line 49: So user can say `environment: local` and it will run in the cloud? Maybe we want to show a warning or change this?

adapter/src/dbt/adapters/fal_experimental/adapter.py
mederka: Yeah, I thought about this, but then I realized that installing dbt-fal is different depending on where isolate is running. This case is especially special because it would happen only if dbt-fal is not installed from pip, but rather locally. And it's OK if isolate is also local, we can just point it to the local dbt-fal directory (as we do in get_default_pip_dependencies), but here isolate is remote. So we either have to upload the local version of dbt-fal or we just install whatever is available from pip. I think that we should definitely print a warning here though. Maybe this is a special enough case to live here?

adapter/src/dbt/adapters/fal_experimental/connections.py
mederka: This is a great suggestion! What do you think, @burkaygur ?

adapter/src/dbt/adapters/fal_experimental/utils/environments.py
mederka: Warning is probably better. We could ask users to use a profile target without a cloud definition to run things locally.

adapter/src/dbt/adapters/fal_experimental/connections.py
mederka: Another thing is that host can also be there if we're using isolate server but not isolate cloud.

chamini2: In that case we could have another object for isolate server? Or what do you think is a good way to handle this?

mederka: I think what we have is OK for now. We can ask our early users for feedback.

burkaygur: I think we shouldnt indent simply because no other adapter does, so I think its good to stay consistent.

chamini2: The only one I see doing this is [BigQuery](https://docs.getdbt.com/reference/warehouse-setups/bigquery-setup#service-account-json). I think you are right.

adapter/src/dbt/adapters/fal_experimental/adapter.py
```diff
@@ -33,29 +40,43 @@ def run_with_adapter(code: str, adapter: BaseAdapter, config: RuntimeConfig) ->
...
+    macro_manifest: MacroManifest,
+    local_packages: Optional[bytes] = None,
+) -> Any:
     # This function can be run in an entirely separate
     # process or an environment, so we need to reconstruct
     # the DB adapter solely from the config.
     adapter = reconstruct_adapter(config, manifest, macro_manifest)
+    fal_scripts_path = get_fal_scripts_path(config)
+    if local_packages is not None:
+        assert not fal_scripts_path.exists()
```
chamini2, line 57: NIT: add an explanation
```suggestion
        assert not fal_scripts_path.exists(), "fal-scripts-path directory exists in environment already"
```

adapter/src/dbt/adapters/fal_experimental/utils/environments.py
```diff
@@ -31,17 +38,31 @@ def run(self, executable: BasicCallable, *args, **kwargs) -> Any:
...
+            # TODO: add docs link
+            logger.warning("`local` environments will be executed on fal cloud." + \
```
chamini2, line 54: great!

adapter/src/dbt/adapters/fal_experimental/utils/environments.py
```diff
@@ -111,14 +149,48 @@ def create_environment(name: str, kind: str, config: Dict[str, Any]):
...
+    else:
+        if config["type"] not in REMOTE_TYPES_DICT.keys():
+            raise RuntimeError(
+                f"Environment type {config['type']} is not supported. Supported types: {REMOTE_TYPES_DICT.keys()}")
+        kind = REMOTE_TYPES_DICT.get(config["type"])
+
+    if 'requirements' not in parsed_config.keys():
+        parsed_config['requirements'] = []
+
+    parsed_config['requirements'].append(f"dill=={importlib_metadata.version('dill')}")
```
chamini2, line 177: ATL: why do we have to add this one manually? 

1. I see we call to `_build_hosted_env` for local fal cloud with no `requirements`. So it would send a `dill==xxx` only requirements. Do we send the found packages in another env and put this one on top? I am a bit confused.

adapter/src/dbt/adapters/fal_experimental/utils/environments.py
```diff
@@ -179,16 +250,29 @@ def _get_dbt_packages(is_teleport: bool = False) -> Iterator[Tuple[str, Optional
...
     dbt_fal_dep = "dbt-fal"
     if _is_pre_release(dbt_fal_version):
-        dbt_fal_path = _get_adapter_root_path()
-        if dbt_fal_path is not None:
-            # Can be a pre-release from PyPI
-            dbt_fal_dep = str(dbt_fal_path)
+        if is_remote:
+            # If it's a pre-release and it's remote, its likely us developing, so we try installing
+            # from Github and we can get the custom branch name from FAL_GITHUB_BRANCH environment variable
+            # TODO: Handle pre-release on PyPI. How should we approach that?
```
chamini2, line 256: I think we may not need to handle this because we want to remove the `_is_pre_release` check for something that assures us we are in dev mode and not just a good guess like that function call. Once we have that, this problem goes away.

adapter/src/dbt/adapters/fal_experimental/utils/environments.py
```diff
@@ -179,16 +250,29 @@ def _get_dbt_packages(is_teleport: bool = False) -> Iterator[Tuple[str, Optional
...
+            import os
+            branch_name = os.environ.get('FAL_GITHUB_BRANCH', 'main')
+
+            dbt_fal_dep = f"git+https://github.com/fal-ai/fal.git@{branch_name}#subdirectory=adapter"
```
chamini2, line 260: Good one!

adapter/src/dbt/adapters/fal_experimental/utils/environments.py
```diff
@@ -179,16 +250,29 @@ def _get_dbt_packages(is_teleport: bool = False) -> Iterator[Tuple[str, Optional
...
+                # Can be a pre-release from PyPI
+                dbt_fal_dep = str(dbt_fal_path)
+                dbt_fal_version = None
 
     dbt_fal_extras = _find_adapter_extras("dbt-fal", is_teleport)
 
     if dbt_fal_extras:
-        dbt_fal_dep += f"[{' ,'.join(dbt_fal_extras)}]"
+        if is_remote:
+            dbt_fal_dep = f"dbt-fal[{' ,'.join(dbt_fal_extras)}] @ {dbt_fal_dep}"
```
chamini2, line 273: wow

chamini2 (approved):
This is great! Thank you! 1 ATL that I'm very interested in.

.github/workflows/test_integration_adapter.yml
```diff
@@ -75,14 +79,21 @@ jobs:
...
+            pip install -e ".[${{ matrix.profile }},teleport,cloud]"
+          elif [[ '${{ matrix.teleport }}' == 'true' ]]
           then
             pip install -e ".[${{ matrix.profile }},teleport]"
+          elif [[ '${{ matrix.cloud }}' == 'true' ]]
+          then
+            pip install -e ".[${{ matrix.profile }},cloud]"
           else
             pip install -e ".[${{ matrix.profile }}]"
           fi
```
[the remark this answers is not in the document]
mederka: If it's OK, I'd like to leave it like this for now.

adapter/src/dbt/adapters/fal_experimental/utils/environments.py
mederka: Yes, locally the user can choose between pickle and dill but for isolate fal they have to use dill. I added it as an extra dependency now.

Also yes, we do layering in `run_in_environment_with_adapter` in adapter.py.

PR #691, opened 2022-12-16
feat: get datawarehouse type in SQL context

chamini2:
## Description
In SQL context we should get the `adapter.type()` to be the encapsulated adpter type, since they process the SQL.
And in all contexts, the target (and env) are the encapsulated adapter type, because they always offer the storage (for Python too).

### Integration tests

Adapter to test:

- postgres
- snowflake
- bigquery
- redshift
- duckdb
- athena

Python version to test:

- 3.8

### DIFF
--- a/adapter/src/dbt/adapters/fal/wrappers.py
+++ b/adapter/src/dbt/adapters/fal/wrappers.py
@@ -22,9 +22,8 @@ class FalCredentialsWrapper:
     def type(self):
         import inspect
 
-        target = filter(lambda s: s.function == "to_target_dict", inspect.stack())
-        db_mat = filter(lambda s: s.function == "db_materialization", inspect.stack())
-        if any(target) or any(db_mat):
+        materializer_funcs = {"to_target_dict", "db_materialization"}
+        if any(frame.function in materializer_funcs for frame in inspect.stack()):
             # This makes sense for both SQL and Python because the target is always the db
             return self._db_creds.type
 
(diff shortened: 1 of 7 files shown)


adapter/src/dbt/adapters/fal/wrappers.py
```diff
@@ -20,6 +20,14 @@ def __init__(self, db_creds: Credentials):
 
     @property
     def type(self):
+        import inspect
+
+        target = filter(lambda s: s.function == "to_target_dict", inspect.stack())
+        db_mat = filter(lambda s: s.function == "db_materialization", inspect.stack())
+        if any(target) or any(db_mat):
```
isidentical, line 27: Just a small style nit:
```suggestion
        materializer_funcs = {"to_target_dict", "db_materialization"}
        if any(frame.function in materializer_funcs for frame in inspect.stack()):
 ```

File: adapter/src/dbt/adapters/fal/wrappers.py [hunk 2]
```diff
@@ -74,8 +83,14 @@ def db_materialization(self, context, materialization):
...
 
-    @classmethod
-    def type(cls):
+    def type(self):
+        import inspect
+
+        render = filter(lambda s: s.function == "render", inspect.stack())
+        db_mat = filter(lambda s: s.function == "db_materialization", inspect.stack())
+        if any(render) or any(db_mat):
+            return self._db_adapter.type()
```
isidentical, line 92: Same as above
```suggestion
        materializer_funcs = {"to_target_dict", "db_materialization"}
        if any(frame.function in materializer_funcs for frame in inspect.stack()):
 ```

isidentical (approved):
This is crazy!!! Thanks a lot for fixing it and adding a test for it

PR #714, opened 2023-01-10
tests(dbt-fal): add resource-paths to dbt_project.yml to make sure it is picked up by dbt

chamini2:
Inspired by a Slack question.


chamini2:
> are the test failures expected?

Yes! I forgot we accesed the files I moved during integration tests.

mederka (approved):
👍

PR #727, opened 2023-01-25
chore: dbt v1.4 support for fal cli

chamini2:
## Notes 

1. Python 3_11 cannot be supported until a new version of dill is published.
2. dbt<1.4 support had to be dropped because of adapter-specific dependencies
    - `google-cloud-bigquery==3.5.0` is necessary for 1.4 but not compatible with `dbt-bigquery<1.4`
    - `snowflake-connector-python` had similar issues
    - Added a dbt-core dependency explicitly about it
    - Left version-sensitive code in case we get requests and **have** to figure out a way to support more versions
      - Probably could make it work for non-bigquery adapters
3. We really need to get the integration tests in order

*****

Adapters to test
postgres, bigquery, snowflake, redshift, duckdb, athena, fal

Python versions
3.7, 3.10


turbo1912 (approved):

chamini2:
got integration tests to pass, reviewing pytest now!

chamini2:
##Notes 

1. Python 3.11 cannot be supported until a new version of dill is published.
2. dbt<1.4 support had to be dropped because of adapter-specific dependencies
    - `google-cloud-bigquery==3.5.0` is necessary for 1.4 but not compatible with `dbt-bigquery<1.4`
    - `snowflake-connector-python` had similar issues

.github/workflows/test_cli.yml
```diff
@@ -15,12 +15,14 @@ jobs:
...
-        python: ["3.7", "3.8", "3.9", "3.10"]
-        dbt: ["1.0.*", "1.1.*", "1.2.*"]
-        exclude:
-          # Only DBT 1.1+ supports Python 3.10 (officially)
-          - python: 3.10
-            dbt: "1.0.*"
+        python:
+          - "3.7"
+          # - "3.8"
+          # - "3.9"
```
mederka, line 21: We should uncomment these?

.github/workflows/test_integration_adapter.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           python -m venv .venv
           source .venv/bin/activate
           pip install --upgrade pip
-          pip install build wheel 'fal[${{ matrix.profile }}]'
+          pip install build wheel '../../fal[${{ matrix.profile }}]'
```
mederka, line 78: Why don't we change the `working-directory` instead?

.github/workflows/test_integration_cli.yml
```diff
@@ -43,6 +43,7 @@ on:
           - "3.8"
           - "3.9"
           - "3.10"
+          - "3.11"
```
mederka, line 46: I thought 3.11 was not supported.

.github/workflows/test_integration_cli.yml
```diff
@@ -123,6 +125,7 @@ jobs:
             "3.8",
             "3.9",
             "3.10",
+            "3.11",
```
mederka, line 128: Remove?

.github/workflows/test_integration_cli.yml
```diff
@@ -215,6 +221,8 @@ jobs:
             dbt: "1.2.*"
           - profile: "athena"
             dbt: "1.3.*"
+          - profile: "athena"
+            dbt: "1.4.*"
```
mederka, line 225: Are we installing `dbt-athena-community`?

projects/adapter/src/dbt/adapters/fal_experimental/utils/environments.py
```diff
@@ -311,15 +310,17 @@ def _version_is_prerelease(raw_version: str) -> bool:
     return package_version.is_prerelease
 
 
-def _get_adapter_root_path() -> Optional[Path]:
-    import dbt.adapters.fal as adapter
+def _get_project_root_path(pacakge: str) -> Path:
+    import fal
```
mederka, line 314: Why is this PR updating the adapter code?

projects/fal/src/fal/fal_script.py
```diff
@@ -13,7 +13,12 @@
 from dbt.config.runtime import RuntimeConfig
 from faldbt.logger import LOGGER
 
-from dbt.contracts.graph.parsed import ColumnInfo
+import faldbt.version as version
+
+if version.is_version_plus("1.4.0"):
+    from dbt.contracts.graph.nodes import ColumnInfo
+else:
+    from dbt.contracts.graph.parsed import ColumnInfo
```
mederka, line 21: Aren't we dropping <1.4?

projects/fal/src/faldbt/logger.py
```diff
@@ -1,101 +1,90 @@
...
+def _prepare_msg(msg: str, *args, **kwargs):
+    if args or kwargs:
+        return msg.format(*args, **kwargs)
+    else:
+        return msg
 
-    def message(self) -> str:
-        return self.msg
+LOGGER = FalLogger()
+log_manager = LogManager(dbt_log_manager)
```
mederka, line 90: Great!

mederka (changes requested):
Great job, @chamini2! I just found some inconsistencies. Open to discussion though.

.github/workflows/test_cli.yml
chamini2: I did this on purpose because if it works for 3.7 and 3.10, it probably works for the ones in the middle. Would you agree?

.github/workflows/test_integration_adapter.yml
chamini2: Because we use the `.venv` in another step in the same working directory as this one.

.github/workflows/test_integration_cli.yml
chamini2: It isn't supported yet, but I am leaving open the option to easily test it.

.github/workflows/test_integration_cli.yml
chamini2: It isn't supported yet, but I am leaving open the option to easily test it.

.github/workflows/test_integration_cli.yml
chamini2: No, this change can be done in a different PR.

projects/adapter/src/dbt/adapters/fal_experimental/utils/environments.py
chamini2: This is just a bug being fixed. Let me take it out to another PR.

projects/fal/src/fal/fal_script.py
chamini2: Yes, but 

> - Left version-sensitive code in case we get requests and have to figure out a way to support more versions
>   - Probably could make it work for non-bigquery adapters

.github/workflows/test_cli.yml
mederka: 3.8 and 3.9 are probably the most popular versions right now, so we should definitely test on those.

mederka: Maybe at least 3.8.

projects/fal/src/fal/fal_script.py
mederka: Right!

mederka (approved):
Looks great, @chamini2! Any chance we get those test to pass?

chamini2:
The errors are becuase of isolate cloud changes. We can address them in a separate PR.

PR #739, opened 2023-02-02
refactor: reorg monorepo s01

drochetti:
## Description

Major repository reorg so it follows the structure and best practices for fal monorepos.

PR #742, opened 2023-02-03
refactor: move some hacks location

chamini2:
FYI: the quoting was fixed in dbt for 1.4+ in https://github.com/dbt-labs/dbt-core/pull/6620

### DIFF
--- a/adapter/src/dbt/adapters/fal/wrappers.py
+++ b/adapter/src/dbt/adapters/fal/wrappers.py
@@ -94,6 +94,15 @@ class FalEncAdapterWrapper(FalAdapterMixin):
         else:
             getattr(super(), name)
 
+    def get_relation(self, database: str, schema: str, identifier: str):
+        # HACK: When compiling Python models, we get an all-False quoting policy
+        #       This does not happen in 1.4
+        if self._db_adapter.type() == "athena":
+            # and dbt-athena-community breaks for that case
+            self.config.quoting = {"database": True, "schema": True, "identifier": True}
+
+        return self._db_adapter.get_relation(database, schema, identifier)
+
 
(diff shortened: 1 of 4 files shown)


drochetti (approved):
should we merge this before #739 ?

adapter/src/dbt/adapters/fal/wrappers.py
```diff
@@ -94,6 +94,15 @@ def __getattr__(self, name):
...
 
+    def get_relation(self, database: str, schema: str, identifier: str):
+        # HACK: When compiling Python models, we get an all-False quoting policy
+        #       This does not happen in 1.4
+        if self._db_adapter.type() == "athena":
+            # and dbt-athena-community breaks for that case
+            self.config.quoting = {"database": True, "schema": True, "identifier": True}
+
+        return self._db_adapter.get_relation(database, schema, identifier)
+
```
mederka, line 105: NIT: Does it matter if we put it in a wrapper or in `impl.py`? I think `impl.py` is better because it's more visible there. This feels more like putting something under a rug.

mederka (approved):

mederka:
> should we merge this before #739 ?

Great suggestion, but this is not going into `main`. It's going to be released as a patch to 1.3.

adapter/src/dbt/adapters/fal/wrappers.py
chamini2: The way I saw this change was taking it from the "general" soluton of the Adapter Mixin and putting it in the wrapper that is actually being used, since this is 1.3-specific and will not be replicated for 1.4

chamini2: merging and you can make a decision in your PR

PR #743, opened 2023-02-03
feat: add adapter_type to telemetry

drochetti:
### Description

Adds a new property to telemetry, the `adapter_type`.

After a bit of trial and error, it seems like the correct (i.e. runtime resolved) field comes from `config.to_target_dict()["type"]`. The more obvious `config.get_metadata().adapter_type` returns the static value from the configuration.

---

**Note:** this PR replaces https://github.com/fal-ai/fal/pull/737


chamini2 (approved):
Let's see tests passing and good to go!

PR #756, opened 2023-02-10
docs: Try to fix PyPI relative links with HTML base tag

chamini2:
An example can be found in https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_base_test

There is a discussion on [pypa](https://github.com/pypa/readme_renderer/issues/163) from 2020 about this.


mederka (approved):
👍 

I guess integration tests are failing because we need to fix koldstart integration.

chamini2:
> 👍
> 
> I guess integration tests are failing because we need to fix koldstart integration.

yes

PR #758, opened 2023-02-14
feat(dbt-fal): `cloud` host resolution

mederka:
## Description
Resolve `host=cloud` to `api.alpha.fal.ai`

### Integration tests

Adapter to test:
- postgres
- bigquery

Python version to test:
- 3.9


chamini2 (approved):
We said to also add a `cloud-eu` case and good to go

projects/adapter/src/dbt/adapters/fal_experimental/connections.py
```diff
@@ -54,6 +60,10 @@ class FalCredentials(Credentials):
     database: str = ""
     schema: str = ""
 
+    def __post_init__(self):
+        if self.host in list(DEFAULT_HOSTS.keys()):
```
chamini2, line 64: this also worked:
```suggestion
        if self.host in DEFAULT_HOSTS:

PR #763, opened 2023-02-16
feat(dbt-fal): replace isolate with koldstart

mederka:


projects/adapter/src/dbt/adapters/fal_experimental/adapter.py
```diff
@@ -80,136 +76,65 @@ def run_in_environment_with_adapter(
...
-
-        extra_config = {
-            'kind': 'virtualenv',
-            'configuration': { 'requirements': deps }
-        }
+    compressed_local_packages = None
 
-        environment.target_environments.append(extra_config)
+    deps = get_default_pip_dependencies(
+        is_remote=(type(environment.host)==KoldstartHost),
```
chamini2, line 82: NIT: you are doing `type ... is` for the same check a few lines below. Maybe store an `is_remote` variable and use it everywhere?

projects/adapter/src/dbt/adapters/fal_experimental/adapter.py
```diff
@@ -80,136 +76,65 @@ def run_in_environment_with_adapter(
...
+    if environment.kind == "virtualenv":
+        requirements = environment.config.get("requirements", [])
+        requirements += deps
         isolated_function = isolated(
             kind="virtualenv",
-            host=host,
-            requirements=requirements,
-            machine_type=environment.machine_type)(execute_model)
-    elif environment.target_env_type == "conda":
+            host=environment.host,
```
chamini2, line 115: this is awesome

projects/adapter/src/dbt/adapters/fal_experimental/adapter.py
```diff
@@ -80,136 +76,65 @@ def run_in_environment_with_adapter(
...
+            env_dict=env_dict)(execute_model)
     else:
         # We should not reach this point, because environment types are validated when the
         # environment objects are created (in utils/environments.py).
-        raise Exception(f"Environment type not supported: {environment.target_env_type}")
+        raise Exception(f"Environment type not supported: {environment.kind}")
+
+    # Machine type is only applicable in KoldstartHost
+    if type(environment.host) is KoldstartHost:
+        isolated_function = isolated_function.on(machine_type=environment.machine_type)
```
chamini2, line 137: ATL: does the isolated function error out if passed this and `LocalHost`?

projects/adapter/src/dbt/adapters/fal_experimental/utils/environments.py
```diff
@@ -39,42 +37,41 @@ def run(self, executable: BasicCallable, *args, **kwargs) -> Any:
...
     # since it will mirror user's execution context directly.
     if environment_name == "local":
         if credentials.key_secret and credentials.key_id:
-            # "local" environment in this case is fal cloud
-            # TODO: add docs link
             logger.warning(
                 "`local` environments will be executed on fal cloud."
                 + "If you don't want to use fal cloud, you can change your "
                 + "profile target to one where fal doesn't have credentials"
             )
```
chamini2, line 63: what do you think of actually deciding this not on credentials but on `host` being there?

chamini2: since you recently added the `host: cloud`, I think it would make sense

projects/adapter/src/dbt/adapters/fal_experimental/utils/environments.py
```diff
@@ -137,81 +133,30 @@ def create_environment(
...
-        "conda": CondaEnvironment,
-        "venv": VirtualPythonEnvironment,
-        "remote": IsolateServer,
-    }
-
-    env_type = REGISTERED_ENVIRONMENTS.get(kind)
-
-    if env_type is None:
+) -> EnvironmentDefinition:
+    if kind not in ["venv", "conda"]:
```
chamini2, line 137: should we support `virtualenv` too? Since that is the remote version?

projects/adapter/src/dbt/adapters/fal_experimental/adapter.py
mederka: `isolated` tries to pass it as a config to the underlining environment, virtualenv or coda, and those don't expect `machine_type` on init. We should work on making config parsing much smarter in koldstart.

projects/adapter/src/dbt/adapters/fal_experimental/utils/environments.py
mederka: `venv` is `virtualenv`, for some reason we rename in `isolate`. Should come to a decision about sticking to one and ditching the other.

projects/adapter/src/dbt/adapters/fal_experimental/utils/environments.py
```diff
@@ -55,7 +55,7 @@ def fetch_environment(
     # Local is a special environment where it doesn't need to be defined
     # since it will mirror user's execution context directly.
     if environment_name == "local":
-        if credentials.key_secret and credentials.key_id:
+        if credentials.host in ["cloud", "cloud-eu"]:
```
chamini2, line 58: I thought we would just check for it being `is not None`? I am not sure how this behaves now

PR #764, opened 2023-02-16
fix(dbt-fal): teleport integration

mederka:


chamini2 (approved):
Thanks for updating teleport!

PR #777, opened 2023-03-10
fix(fal): handling the logger in isolated scripts and models

mederka:


projects/fal/src/faldbt/logger.py
```diff
@@ -17,6 +17,17 @@ def __init__(self):
...
 
+    def __getstate__(self):
+        # Don't pickle the logger
+        d = self.__dict__.copy()
+        d['_logger'] = d['_logger'].name
+        return d
+
+    def __setstate__(self, d):
+        # Set logger when unpickling
+        d['_logger'] = logging.Logger("fal", logging.INFO)
```
chamini2, line 28: If you are saving the `.name` of the logger while pickling, would it make sense to do
```suggestion
        d['_logger'] = logging.Logger(d['_logger'], logging.INFO)
```
instead?

chamini2: Maybe

```py
d['_logger'] = logging.Logger(d['_logger'] if isinstance(d['_logger'], str) else "fal", logging.INFO)
```

chamini2 (approved):

chamini2 (approved):

PR #779, opened 2023-03-13
feat(dbt-fal): make dbt-fal work with sqlserver

### LINKED ISSUE #778: Support sqlserver
Add support for MS sqlserver

mederka:
## Description
Adds support for sqlserver adapter
Closes #778

### DIFF
--- a/.github/workflows/test_integration_adapter.yml
+++ b/.github/workflows/test_integration_adapter.yml
@@ -32,6 +32,7 @@ jobs:
           - snowflake
           - redshift
           - trino
+          # - sqlserver # does not support dbt 1.4 yet
         dbt_version:
           - "1.4.*"
         python:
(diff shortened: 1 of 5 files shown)


mederka:
@chamini2, should we add integration tests for sqlserver? Not sure about the overhead.

chamini2:
> @chamini2, should we add integration tests for sqlserver? Not sure about the overhead.

I would say we should @mederka. We can later make some tests run only in certain cases, but having the integration tests seems important if we have sqlserver code

.github/workflows/test_integration_adapter.yml
```diff
@@ -32,6 +32,7 @@ jobs:
           - snowflake
           - redshift
           - trino
+          # - sqlserver # does not support dbt 1.4 yet
```
chamini2, line 35: OK, this is interesting for me. We deploy latest from this branch. Should this be pointing to a [1.3 branch](https://github.com/fal-ai/fal/tree/release-adapter-1.3)?

chamini2: Or just wait for 1.4?

mederka: I think we can just wait.

chamini2 (approved):

PR #780, opened 2023-03-13
feat: send fal script telemetry info

chamini2:
## Description
global and hook information on scripts.

### Integration tests

Adapter to test:

- postgres

Python version to test:

- 3.8

### DIFF
--- a/projects/fal/src/faldbt/project.py
+++ b/projects/fal/src/faldbt/project.py
@@ -512,7 +512,6 @@ class FalDbt:
 
         telemetry.log_api(
             action="faldbt_initialized",
-            additional_props={"config_hash": self._config.hashed_name()},
             dbt_config=self._config,
         )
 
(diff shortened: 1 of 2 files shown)


projects/fal/src/faldbt/project.py
```diff
@@ -512,7 +512,6 @@ def __init__(
 
         telemetry.log_api(
             action="faldbt_initialized",
-            additional_props={"config_hash": self._config.hashed_name()},
```
mederka, line 515: Why are we deleting this?

chamini2: This is being set inside `log_api` already, results in no functionality change

https://github.com/fal-ai/fal/blob/3e955628aed3ec728b2d644b84727d3677918f4e/projects/fal/src/fal/telemetry/telemetry.py#L297-L298

mederka (approved):

PR #788, opened 2023-03-14
feat(dbt-fal): add fal cloud context

mederka:
Provides new cloud context, with `store_open` method that works like `open` but in cloud context.

```
def model(dbt, fal):
    from utils.get_bool import get_bool
    with fal.store_open("test", "w") as f:
        f.write("hello world")

    df = dbt.ref("model_b")

    df["my_bool"] = get_bool()
    return df
```

### DIFF
--- a/projects/adapter/src/dbt/adapters/fal_experimental/adapter.py
+++ b/projects/adapter/src/dbt/adapters/fal_experimental/adapter.py
@@ -19,6 +19,7 @@ from dbt.adapters.fal_experimental.utils.environments import (
 from dbt.parser.manifest import MacroManifest, Manifest
 
 from .adapter_support import (
+    FalCloudContext,
     prepare_for_adapter,
     read_relation_as_df,
     reconstruct_adapter,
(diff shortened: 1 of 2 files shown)


projects/adapter/src/dbt/adapters/fal_experimental/adapter.py
```diff
@@ -37,6 +38,7 @@ def run_with_adapter(code: str, adapter: BaseAdapter, config: RuntimeConfig) ->
         return main(
             read_df=prepare_for_adapter(adapter, read_relation_as_df),
             write_df=prepare_for_adapter(adapter, write_df_to_relation),
+            fal_context=FalCloudContext() if is_cloud else None
```
chamini2, line 41: I would say the context should always be there, not just for cloud.

```py
class FalContext:
  ...

class FalCloudContext(FalContext):
  ... # implement the differences

class FalLocalContext(FalContext):
  ...
```

chamini2: We always thought of these things working the same locally and in the cloud.

projects/adapter/src/dbt/adapters/fal_experimental/adapter_support.py
```diff
@@ -208,3 +209,24 @@ def reload_adapter_cache(adapter: BaseAdapter, manifest: Manifest) -> None:
...
         yield adapter.connections.get_thread_connection()
+
+
+class FalCloudWriter(object):
+    def __init__(self, path: str, options: str):
+        self.path = path
+        self.options = options
+
+    def __enter__(self):
+        self.file = open(self.path, self.options)
```
chamini2, line 220: ATL: is this opening in the `/data` dir or a temp one in the cloud?

PR #797, opened 2023-03-28
docs: add serverless stable diffusion example

mederka:


docsite/docs/fal-serverless/examples/stable_diffusion.md
```diff
@@ -2,6 +2,88 @@
...
+    from diffusers import StableDiffusionPipeline, DPMSolverMultistepScheduler
+
+    model_id = "runwayml/stable-diffusion-v1-5"
+    os.environ['TRANSFORMERS_CACHE'] = '/data/hugging_face_cache'
+
+    pipe = StableDiffusionPipeline.from_pretrained(
+        model_id,
+        num_inference_steps=20,
+        torch_dtype=torch.float16,
+        cache_dir="/data/hfcache")
```
burkaygur, line 58: is this supposed to be os.environ['TRANSFORMERS_CACHE'] ?

docsite/docs/fal-serverless/examples/stable_diffusion.md
```diff
@@ -2,6 +2,88 @@
...
+
+    generator = torch.Generator("cuda")
+    image = pipe(prompt, generator=generator).images[0]
+
+    buf = io.BytesIO()
+    image.save(buf, format="PNG")
+    return buf
+```
+We set the package and machine type requirements as `isolated` argument. `generate` function then imports the necessary modules and initializes the stable diffusion pipeline using the specified model ID. It also sets up the Hugging Face transformers cache directory to use fal-serverless persistent [/data directory](../storage_and_persistence). The pipeline is then moved to the GPU and configured with a custom scheduler. Finally, the function generates the image based on the given prompt and returns it as a byte buffer.
+
```
burkaygur, line 70: We might want to add a section here to introduce setup functions too?

mederka: I'm just going to link to reference docs. I don't think that this example needs to be overly complex.

PR #802, opened 2023-03-30
docs: add workflow dispatch inputs

mederka:


chamini2:
is this still valid?

mederka:
> is this still valid?

Hi @chamini2, yes! It would be great if we could deploy docs with workflow dispatch.

chamini2 (approved):

chamini2:
Sounds gret

PR #814, opened 2023-04-12
feat(dbt-fal): adapt to new 1.5 dbt changes

chamini2:
## Description
Refactor and adapt DB profile loading for dbt-fal.
It mostly follows the same ideas adapted to some changes on how flags are loaded and avoiding a new parser for some checks.
In general this is a better implementation than before but does not really take advantage of any new API introduced in 1.5

Other good thing with this change is that now running `dbt run` on the integration test projects does not need a `temp_dir` env var being set, becuase it handles it exclusively for behave.

## Postgres
Got it to just fail for `features/source.feature:9  Run a Python model that queries a source in local environment`.
That failure would be related to fal-cli tool, because we fill the source with a `FalDbt` instance.

```
❯ python load_freshness_table.py . ~/.dbt/
  File "load_freshness_table.py", line 3, in <module>
    from fal import FalDbt
  File "/Users/matteo/.pyenv/versions/fal15/lib/python3.8/site-packages/fal/__init__.py", line 3, in <module>
    from faldbt.project import (
  File "/Users/matteo/.pyenv/versions/fal15/lib/python3.8/site-packages/faldbt/project.py", line 20, in <module>
    from dbt.contracts.graph.parsed import (
ModuleNotFoundError: No module named 'dbt.contracts.graph.parsed'
```

## Redshift
Changed connection solution, will be addressed in PR #835

### DIFF
--- a/projects/adapter/integration_tests/features/steps/fal_adapter_steps.py
+++ b/projects/adapter/integration_tests/features/steps/fal_adapter_steps.py
@@ -10,6 +10,10 @@ from datetime import datetime
 import re
 
 
+def target_path(context):
+    return str(Path(context.temp_dir.name) / "target")
+
+
 @when("the following shell command is invoked")
 def run_command_step(context):
     context.exc = None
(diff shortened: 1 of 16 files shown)


projects/adapter/integration_tests/features/steps/fal_adapter_steps.py
```diff
@@ -46,7 +50,7 @@ def set_project_folder(context, project: str):
     context.base_dir = str(project_path)
     context.temp_dir = tempfile.TemporaryDirectory()
     context.project_name = _load_dbt_project_file(context)["name"]
-    os.environ["temp_dir"] = context.temp_dir.name
+    os.environ["DBT_TARGET_PATH"] = target_path(context)
```
mederka, line 53: 👍

mederka (approved):
Great job, @chamini2!

PR #829, opened 2023-04-20
fix(fal): profiles without default targets

mederka:


mederka:
> awesome, maybe we can "test" this sort of stuff by changing some profiles.yml to use default (was that the repro case?)

That's a great idea!

PR #835, opened 2023-04-26
fix(dbt-fal): adapt to new redshift connector library for 1.5

chamini2:
closes FEA-1014


mederka (commented):
This looks good! Does it work?

chamini2:
It doesn't work yet, we have to fix how the connection is actually used. I did this based on something I found online but it have not tested locally yet.

PR #841, opened 2023-04-28
feat(fal): add dbt 1.5 support 

### LINKED ISSUE #846: fal is not compatible with dbt 1.5.0
We upgraded the dbt version to 1.5.0, but when this upgrade finished it says that the version of dbt is not compatible with fal (0.8.6)

chamini2:
- snowflake
- bigquery
- postgres

*****

https://github.com/dbt-labs/dbt-core/issues/7465 is making it hard to reliably know where to search for artifacts (e.g. run_results) so we are still using the deprecated target-path in dbt_project.yml for pytest and we may need to use it for integration tests

closes #846 
closes FEA-1037

### DIFF
--- a/projects/fal/integration_tests/projects/000_fal_run/fal_scripts/after.py
+++ b/projects/fal/integration_tests/projects/000_fal_run/fal_scripts/after.py
@@ -18,7 +18,7 @@ output = output + f"\nModel dataframe information:\n{info}"
 output = output + f"\nModel alias is {model_alias}"
 alias_no_namespace = model_alias.split('__ns__')[1]
 output = output + f"\nModel alias without namespace is {alias_no_namespace}"
-temp_dir = os.environ["temp_dir"]
+temp_dir = os.getenv("temp_dir", ".")
 write_dir = open(reduce(os.path.join, [temp_dir, model_name + ".after.txt"]), "w")
 write_dir.write(output)
 write_dir.close()
--- a/projects/fal/src/fal/cli/flow_runner.py
+++ b/projects/fal/src/fal/cli/flow_runner.py
@@ -47,10 +47,8 @@ def run_threaded(
 
 
 def fal_flow_run(parsed: argparse.Namespace) -> int:
-    arg_vars = getattr(parsed, "vars", "{}")
-
     # fal-format Python models
-    generated_models = generate_python_dbt_models(parsed.project_dir, arg_vars)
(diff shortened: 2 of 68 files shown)


projects/fal/integration_tests/projects/000_fal_run/fal_scripts/after.py
```diff
@@ -18,7 +18,7 @@
 output = output + f"\nModel alias is {model_alias}"
 alias_no_namespace = model_alias.split('__ns__')[1]
 output = output + f"\nModel alias without namespace is {alias_no_namespace}"
-temp_dir = os.environ["temp_dir"]
+temp_dir = os.getenv("temp_dir", ".")
```
mederka, line 21: 👍

projects/fal/src/fal/cli/flow_runner.py
```diff
@@ -47,10 +47,8 @@ def run_threaded(
 
 
 def fal_flow_run(parsed: argparse.Namespace) -> int:
-    arg_vars = getattr(parsed, "vars", "{}")
-
     # fal-format Python models
-    generated_models = generate_python_dbt_models(parsed.project_dir, arg_vars)
+    generated_models = generate_python_dbt_models(parsed.project_dir, parsed.vars)
```
mederka, line 51: When are we dropping these?

projects/fal/src/faldbt/parse.py
```diff
@@ -89,6 +89,13 @@ def get_dbt_config(
         os.chdir(owd)
     else:
         config = RuntimeConfig.from_args(args)
+
+    # HACK: issue in dbt-core 1.5.0 https://github.com/dbt-labs/dbt-core/issues/7465
+    env_target_path = os.getenv("DBT_TARGET_PATH")
+    if env_target_path:
+        config.target_path = env_target_path
```
mederka, line 96: Will target changes require any action from current users that want to update to 1.5?

mederka (approved):
I think this looks great! Really awesome how things just seem to be tidier.

Great job, @chamini2 👏

projects/fal/src/fal/cli/flow_runner.py
chamini2: > When are we dropping these?

Did not really underrstand the question.

I just introduced `--vars` support for `fal run` too (wanted to use vars for a different use and added it). But it's really not related to dbt 1.5

PR #845, opened 2023-05-04
fix(adapter): make dbt-fal~1.5 work with redshift

mederka:


projects/adapter/pyproject.toml
```diff
@@ -54,7 +55,7 @@ pytest = "^5.2"
 postgres = []
 snowflake = ["snowflake-connector-python"]
 bigquery = ["google-cloud-bigquery"]
-redshift = ["sqlalchemy-redshift"]
+redshift = ["sqlalchemy-redshift", "awswrangler"]
```
chamini2, line 58: do we still need `sqlalchemy-redshift`?

chamini2 (commented):
Looks pretty good!

projects/adapter/pyproject.toml
```diff
@@ -34,6 +34,7 @@ google-cloud-bigquery = { version = "~3.5.0", extras = ["pandas"], optional = tr
 
 ## redshift
 sqlalchemy-redshift = { version = "^0.8.9", optional = true }
+awswrangler = { version = ">=3.0.0", extras = ["redshift"], optional = true}
```
chamini2, line 37: also curious if you considered the one used by dbt-redshift?

https://github.com/dbt-labs/dbt-redshift/blob/b644963a6bf7d3e64bd71e6ee97a205900089db1/setup.py#L87-L88

projects/adapter/pyproject.toml
mederka: Good catch! I don't think so.

projects/adapter/pyproject.toml
mederka: Yeah, thank you for suggesting, @chamini2! awswrangler just seemed tidier. They also have pretty good Athena support, so I was thinking that maybe we use it for Athena as well.

mederka: https://github.com/aws/aws-sdk-pandas

chamini2: Very interesting! Let's try that!

chamini2 (approved):
Awesome!

PR #850, opened 2023-05-12
fix: load flags correctly for dbt loading to work with dbt-fal

chamini2:


projects/fal/src/faldbt/lib.py
```diff
@@ -84,27 +105,13 @@ def _get_adapter(
...
-
-    try:
+    with adapter.cache.lock:
         yield
-    except:
-        LOGGER.debug("Error during lock operation {}", operationId)
-        raise
-    finally:
-        _lock.release()
-        LOGGER.debug("Released {}", operationId)
```
chamini2, line 107: Point of interest: this `RLock` when acquired in an isolated environment and with the `dbt-fal` adapter, it would not acquire it ever, instead it would hang in `_lock.acquire()` forever.

I am not sure what caused it and changed it to rely on `adapter.cache.lock` for now.

Any ideas on why this may have been? I don't love relying on the internal adapter lock.

chamini2: @isidentical found the reason! https://github.com/uqfoundation/dill/issues/321

chamini2: reverted by using `threding._PyRLock`

projects/fal/src/faldbt/project.py
```diff
@@ -456,23 +457,42 @@ def __init__(
...
+
+        vars = parse.parse_cli_vars(args_vars)
+
+        flags = lib.initialize_dbt_flags(
+            profiles_dir=profiles_dir,
+            project_dir=project_dir,
+            threads=threads,
+            profile_target=profile_target,
+            vars=vars,
+        )
```
mederka, line 477: Nice

mederka (approved):
Looks good, @chamini2

PR #851, opened 2023-05-15
docs: fix example file path

kiwamizamurai:
## Description

 fix example file path

### Integration tests

nothing


chamini2 (approved):
Thanks for catching this!

Issue #858, opened 2023-05-27
Case sensitive naming leads to a "When searching for a relation, dbt found an approximate match" error


flunardelli:
**Describe the bug**
Using sql server adapter with a database.schema.table in PascalCase leads a relationship error when I try rerun a dbt-fal python model.

**Your environment**
- OS: MacOS / Windows / Linux 
```
Completed with 1 error and 0 warnings:
  
20:00:58.046485 [debug] [MainThread]: Command end result
20:00:58.056627 [info ] [MainThread]: 
20:00:58.056918 [info ] [MainThread]: Completed with 1 error and 0 warnings:
20:00:58.057071 [info ] [MainThread]: 
20:00:58.057227 [error] [MainThread]: Compilation Error in model Int_Base_FalTest (models/intermediate/Base/Int_Base_FalTest.py)
20:00:58.057377 [error] [MainThread]:   When searching for a relation, dbt found an approximate match. Instead of guessing 
20:00:58.057500 [error] [MainThread]:   which relation to use, dbt will move on. Please delete "AdventureWorksDW2019"."EDWStaging"."Int_Base_FalTest", or rename it to be less ambiguous.
20:00:58.057635 [error] [MainThread]:   Searched for: "adventureworksdw2019"."edwstaging"."int_base_faltest"
20:00:58.057753 [error] [MainThread]:   Found: "AdventureWorksDW2019"."EDWStaging"."Int_Base_FalTest"
20:00:58.057881 [error] [MainThread]:   
20:00:58.057996 [error] [MainThread]:   > in macro is_incremental (macros/materializations/models/incremental/is_incremental.sql)
20:00:58.058124 [error] [MainThread]:   > called by macro py_script_postfix (macros/python_model/python.sql)
20:00:58.058248 [error] [MainThread]:   > called by model Int_Base_FalTest (models/intermediate/Base/Int_Base_FalTest.py)
20:00:58.058386 [info ] [MainThread]: 
20:00:58.058520 [info ] [MainThread]: Done. PASS=0 WARN=0 ERROR=1 SKIP=0 TOTAL=1
20:00:58.058712 [debug] [MainThread]: Flushing usage events

dbt-fal==1.4.8
dbt-core==1.4.6
```
- Adapter being used:  
dbt-sqlserver==1.4.3
**How to reproduce**
Create an sql server model using PascalCase naming. Run it once and try update it after that.

**Expected behavior**
A model been updated without this relation search error

**Actual behavior**
Compilation Error

[...]

PR #859, opened 2023-05-31
fix(dbt-fal): sqlserver quoting policy

mederka:
Closes: #858


chamini2 (approved):
Thanks!

chamini2:
it seems env tests are failing 

> Failing scenarios:
  features/environments.feature:5  Run a project with a Python model that is calculated in a virtual environment
  features/source.feature:9  Run a Python model that queries a source in local environment

Maybe it's fine for now

PR #860, opened 2023-06-01
fix(dbt-fal): update datadog-api-client dependency

mederka:


chamini2:
This does not really solve the issue, since 2.13.0 introduced the error

chamini2:
closing as #862 was merged

PR #882, opened 2023-07-10
chore: merge dbt-fal and fal packages into dbt-fal

chamini2:
The idea of this PR is to merge the 2 packages into one, and we would start releasing only the dbt-fal package.

# New `dbt-fal` command line tool

We now install the fal and dbt-fal CLIs, they are the same but the `fal` one gets a warning when started.

```
❯ fal flow run
The dbt tool `fal` and `dbt-fal` adapter have been merged into a single tool.
Please use the new `dbt-fal` command line tool instead.
Running `pip install dbt-fal` will install the new tool and the adapter alongside.
Then run your command like

    dbt-fal <command>

00:29:04  Running with dbt=1.5.1
00:29:04  Unable to do partial parsing because config vars, config profile, or config target have changed
...

❯ dbt-fal flow run
00:29:10  Running with dbt=1.5.1
00:29:10  Unable to do partial parsing because config vars, config profile, or config target have changed
...
```

# Warning on fal imports

Another thing to consider is that now imports from `fal` moved to `fal.dbt`.  Now we get warnings for `fal` direct imports.

```sh
>>> import fal
>>> fal.FalDbt
The dbt tool `fal` and `dbt-fal` adapter have been merged into a single tool.
Please import from the `fal.dbt` module instead.
Running `pip install dbt-fal` will install the new tool and the adapter alongside.
Then import from the `fal.dbt` module like

    from fal.dbt import FalDbt

<class 'fal.dbt.utils.FalDbt'>
>>> fal.FalDbt # no more warnings
<class 'fal.dbt.utils.FalDbt'>
>>> from fal import NodeStatus
The dbt tool `fal` and `dbt-fal` adapter have been merged into a single tool.
Please import from the `fal.dbt` module instead.
Running `pip install dbt-fal` will install the new tool and the adapter alongside.
Then import from the `fal.dbt` module like

    from fal.dbt import NodeStatus

>>> from fal import NodeStatus # no more warnings
>>> from fal.dbt import Context
```

To later free the fal namespace.

*****

Run tests for:

- postgres
- bigquery
- snowflake

*****

Closes FEA-1227
Closes FEA-1263
Closes FEA-1267


projects/adapter/cli_tests/projects/002_jaffle_shop/fal_dbt.py
```diff
@@ -2,7 +2,7 @@
 from fal import FalDbt
```
burkaygur, line 2: Is this one going to change?

mederka (approved):
Wow!

This is epic work, @chamini2 !

I have two small questions:

- Some tests are set to @TODO, does this mean that this feature doesn't work for those databases for both CLI and adapter?
- Will a person be able to use fal-serverless (new fal) and dbt-fal? I see that we're putting the fal cli in fal namespace, so just curious if the CLI and serverless functions can work together in the same environment.

chamini2:
> Will a person be able to use fal-serverless (new fal) and dbt-fal? I see that we're putting the fal cli in fal namespace, so just curious if the CLI and serverless functions can work together in the same environment.

Even though fal cli is in the fal namespace, notice that the whole project is in `src/fal/dbt/*`. That means that to import _anything_ from it you use `import fal.dbt`.

This is the same behavior we see for dbt adapters. They install modules in the dbt namespace, but inside the `dbt.adapters.<adapter-name>` module, so it does not conflict with imports from dbt-core .

Let me know if I did not explain myself well and I can clarify!

chamini2:
> Some tests are set to @todo, does this mean that this feature doesn't work for those databases for both CLI and adapter?

The tests right now keep being separate, we test the cli and adapter separately. 

So a todo in the cli_tests folder applies to cli and a todo in the integration_tests folder applies to adapter.

burkaygur:
🚢 thank you @chamini2

PR #889, opened 2023-07-18
feat: add fal-serverless project 

chamini2:
# TODO
- [ ] configure secrets and env variables of repo
- [ ] publish a 0.1 version for isolate-proto
- [ ] should we bring isolate-proto here too?


chamini2:
cancelled for now

PR #895, opened 2023-08-08
fix: add name to DbtGenericTest repr

mederka:


chamini2 (approved):
let's wait for confirmation that `.name` was it

PR #900, opened 2023-08-29
update to dbt-core 1.5.5

dnascimento:
## Description
Add support to dbt-core 1.5.5


chamini2:
Hey @dnascimento , thanks for idea but the internals of dbt-core changed from 1.5.1 to 1.5.2 and there are necessary changes for this to actually work, this is the issue on their repo https://github.com/dbt-labs/dbt-core/issues/7465 which led to changes.

PR #919, opened 2024-02-26
support dbt-core between 1.6.0 to 1.7.8

salimmoulouel:
## Description
resolves  
https://github.com/fal-ai/dbt-fal/issues/909
https://github.com/fal-ai/dbt-fal/issues/880
https://github.com/fal-ai/dbt-fal/issues/914

### Integration tests

Adapter to test:

- bigquery

Python version to test:

- 3.11

Issue #914, opened 2024-01-08
Is dbt-fal abandonware?


candalfigomoro:
I see that dbt-fal is falling far behind on compatibility with new versions of dbt, is it abandonware or still being developed?

alvaro-ponce:
even the documentation page is down... I am afraid I discovered what I need just in time to not use it :(

salimmoulouel:
I tried to adapt it to the new version here is the Merge request, i'm waiting for review of the owner 
https://github.com/fal-ai/dbt-fal/pull/919

alvaro-ponce:
ouch...
thanks @HlebKrah

salimmoulouel:
I require assistance in implementing a test. This marks my inaugural contribution to this repository and only my second contribution in my entire developer journey.

PR #922, opened 2024-04-04
Feature/support dbt core1.7.x

### LINKED ISSUE #909: upgrade to support dbt-core v1.7.0
## Background 

Minor version `v1.7` is targeted for final release on Ocotber 26, 2023. As a maintainer of a dbt adapter, we strongly encourage you to release a corresponding minor version increment to ensure users of your adapter can make use of this new minor version.

## How to upgrade

https://github.com/dbt-labs/dbt-core/discussions/8307 is an open discussion with more detailed information. If you have questions, please put them there!

kudryk:
## Description

This PR is building upon the work @salimmoulouel started with PR #919. There's a fix to limit Google protobuf version which this project doesn't use but dbt-core does; dbt-core is fixing that in version 1.8; as this PR is still focused on 1.7.x I've added it here.

resolves
#909
#880
#914

### Integration tests

Adapter to test:
- postgres

Python version to test:
- 3.8
- 3.9
- 3.10
- 3.11


projects/adapter/tests/test_cli.py
```diff
@@ -252,7 +252,7 @@ def test_selection(capfd):
 
 def test_no_run_results(capfd):
     with ProjectTemporaryDirectory() as tmp_dir:
-        shutil.rmtree(os.path.join(tmp_dir, "mockTarget"))
+        shutil.rmtree(os.path.join(tmp_dir, "mockTarget"), ignore_errors=True)
```
squat, line 255: This seems sketchy! Why do we have to ignore errors?

kudryk: Perhaps I should have deleted the line instead. ProjectTemporaryDirectory() creates the temp directory but without `mockTarget`, yet `rmtree` will try to delete the entire tree assuming `mockTarget` exists, but as it doesn't, an exception is thrown. Unsure why this line was put there in the first place. I added ignore_errors as the rmtree API has been updated; perhaps this is a new python 3.11 behaviour.

kudryk: @squat I've updated this test to check for the directory's existence first.

